Issue #319, opened 2026-06-21
Include OSM changeset and version in output


Cj-Malone:
The pmtiles includes the object id, but not the version, by the time the data is used, OSM could have been updated. Either to add the missing data, or to turn the POI into something else. Including the version would allow tools to skip objects that have been edited.

brawer:
> no need for the changeset

Indeed! https://github.com/alltheplaces/osm-diffs/issues/730

PR #334, opened 2026-06-29
Include OSM changeset and version in output

brawer:
We may not actually need the changeset IDs; thanks to @CJ-Malone for noticing! Keeping them around for now, but will remove them again if they turn out to be unnecessary.

See https://pmtiles.io/#url=https%3A%2F%2Fcdn.diffed-places.org%2Fedits.pmtiles&inspectFeatures=true for the PMTiles file after this change.

https://github.com/alltheplaces/osm-diffs/issues/319

### DIFF
--- a/src/matchers/mod.rs
+++ b/src/matchers/mod.rs
@@ -350,6 +350,9 @@ impl<'a> Matcher for ShopMatcher<'a> {
         }
     }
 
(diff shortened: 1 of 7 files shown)


Cj-Malone (approved):
I can't really review Rust, but the output looks good.

PR #539, opened 2026-08-07
Drop exact-duplicate ring members instead of letting them self-cancel (fixes #532)

brawer:
## What

Fixes #532: `PolygonAssembler`'s even-odd fill rule toggles at every ring boundary it crosses, so a ring added twice -- the same OSM way listed twice as a relation member, or two different ways that happen to trace the same boundary -- toggles fill *twice* at every point along it, canceling back out to "absent". If the duplicated ring was meant to be a hole, the hole silently disappears; if it was the only ring, the whole result comes back empty.

## How

`add_ring()` now canonicalizes each incoming ring (`ring_key`: smallest rotation of its point sequence, tried in both winding directions, so the same boundary keys identically regardless of which point it starts at or which way it's wound) and drops any ring whose key was already seen, before it can reach the even-odd union at all.

## Grid fixture impact

Fixes all 4 of the `grid_tests.rs` `KNOWN_MISMATCHES` cases this affected:
- `790` -- the same way listed twice.
- `791`/`792` -- two different ways tracing the same ring, one of them rotated to a different starting point.
- `795` -- which #531's earlier triage had miscategorized as a stitched-closed-loop case; turns out it's actually the same way listed twice as an `inner` member, so it belongs here instead.

## Testing

- New `PolygonAssembler` tests cover the duplicate-outer, duplicate-hole, and rotated/reversed-duplicate cases directly (mirroring grid `790`, `795`, `792`).
- `cargo test` (full suite) -- green.
- `cargo clippy`, `cargo fmt`, `cargo doc` -- clean.

Remaining `grid_tests.rs` `KNOWN_MISMATCHES` are now only the #537 "spike" cases (`711`, `742`, `743`) and fixtures with no lenient oracle to check against at all.

### DIFF
--- a/src/geometry/polygon_assembler.rs
+++ b/src/geometry/polygon_assembler.rs
@@ -272,13 +272,6 @@ fn extract_rings(geom: &Geometry<f64>) -> Vec<LineString<f64>> {
 /// lexicographically smallest -- O(n²) in the ring's point count, fine
 /// given `PolygonAssembler`'s own coordinate budget bounds how large a
 /// stored ring ever gets.
-///
-/// f64 coordinates are compared by bit pattern (via `to_bits()`) rather
-/// than `PartialOrd`/`PartialEq`, sidestepping NaN's lack of a total
-/// order -- exact bit-identical coordinates are what actually occurs here
-/// (a duplicated OSM way reuses the same underlying node coordinates
-/// verbatim, not a recomputed approximation of them), so this doesn't
-/// give up any real matches.
 fn ring_key(ls: &LineString<f64>) -> Vec<(u64, u64)> {
     let all = &ls.0;
     let n = if all.len() > 1 && all.first() == all.last() {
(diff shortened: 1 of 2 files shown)


src/geometry/polygon_assembler.rs
```diff
@@ -239,6 +264,44 @@ fn extract_rings(geom: &Geometry<f64>) -> Vec<LineString<f64>> {
...
+    let mut points = ls.0.clone();
+    if points.len() > 1 && points.first() == points.last() {
+        points.pop(); // the closing repeat; rotation re-closes it either way
+    }
+    if points.is_empty() {
+        return Vec::new();
+    }
+
+    let bits = |c: &Coord<f64>| (c.x.to_bits(), c.y.to_bits());
+    let mut reversed = points.clone();
```
brawer: Here we're currently constructing a Vec with the points in reversed order. Do we really need this clone, or could we iterate the original points in reversed order, for examle by looking up seq[n - ...], and only do the copy if needed? I'd assume that bits() is so cheap that we can do it twice, but cloning a vector with potentially thousands of members is less cheap?

src/geometry/polygon_assembler.rs
```diff
@@ -280,23 +280,31 @@ fn extract_rings(geom: &Geometry<f64>) -> Vec<LineString<f64>> {
...
-    let n = points.len();
+    let bits = |c: Coord<f64>| (c.x.to_bits(), c.y.to_bits());
+    // `all[j]` walks the ring forward from index 0; `all[n - 1 - j]` walks
+    // it backward from the end -- reading both directions off the same
+    // slice by index, rather than materializing a second, reversed copy
+    // of a ring that can hold thousands of points.
+    let at = |forward: bool, j: usize| if forward { all[j] } else { all[n - 1 - j] };
 
-    [&points, &reversed]
+    [true, false]
```
brawer: This is difficult to understand, especially in combination with the multiple iterators. Perhaps extract some of the logic into a helper method, which would explain what it does?

PR #578, opened 2026-08-10
Fix all clippy warnings

brawer:
## Summary
Fixes clippy warnings across 14 files. No behavior changes — purely lint cleanups plus `cargo fmt`.

This intentionally excludes `src/pipeline/osm/id_tagging_schema.rs`, which is generated by `scripts/generate_id_tagging_schema.py` and still has `bool_assert_comparison` warnings (`assert_eq!(x, true/false)`). Since hand-editing a generated file would just get overwritten on the next regeneration, that fix belongs in the generator script and will be done in a follow-up PR.

## Changes
- `assert_eq!(x, true/false)` → `assert!(x)` / `assert!(!x)` in tests
- Removed needless borrows (`&path.path()`, `&geojson`, etc.)
- Changed `const TEST_X: LazyLock<...>` to `static` (interior mutability in const), which also fixes the downstream `borrow_interior_mutable_const` warnings on those items
- Replaced `x == None` with `x.is_none()`
- Replaced manual `Option::map(|x| x.clone())` with `.cloned()`
- Elided a redundant lifetime in `atp::mod::tags`
- Removed a redundant `use predicates;` import
- Fixed inconsistent digit grouping (e.g. `11_000_000_0` → `110_000_000`)
- Replaced `assert!(false, ...)` with `panic!(...)`
- Replaced `.into_iter()` on an array reference with `.iter()`
- Removed a redundant `&` in a `format!` argument

## Testing
- `cargo clippy --locked --all-targets -- -D warnings` → clean, except for the pre-existing warnings in the generated `id_tagging_schema.rs` (see above)
- `cargo test --locked` → 200 lib tests + 4 integration tests pass
- `cargo fmt` applied


src/coverage.rs
```diff
@@ -40,12 +40,12 @@ mod tests {
 
     #[test]
     fn test_is_wikidata_id() {
-        assert_eq!(is_wikidata_key("highway"), false);
-        assert_eq!(is_wikidata_key("wikidata"), true);
-        assert_eq!(is_wikidata_key("brand:wikidata"), true);
-        assert_eq!(is_wikidata_key("network:wikidata"), true);
-        assert_eq!(is_wikidata_key("operator:wikidata"), true);
-        assert_eq!(is_wikidata_key("species:wikidata"), false);
+        assert!(!is_wikidata_key("highway"));
```
brawer: This is now difficult to read. Split the positive and negative test cases into two groups. Still in the same test case, just visually separated in two blocks.

brawer: Done in 3048ae7 — split into a "true" block and a "false" block.

src/tables/u64_set.rs
```diff
@@ -134,25 +134,25 @@ mod tests {
...
     #[test]
     fn test_contains() {
-        assert_eq!(TEST_TABLE.contains(u64::MIN), false);
-        assert_eq!(TEST_TABLE.contains(7), true);
-        assert_eq!(TEST_TABLE.contains(23), true);
-        assert_eq!(TEST_TABLE.contains(41), false);
-        assert_eq!(TEST_TABLE.contains(42), true);
-        assert_eq!(TEST_TABLE.contains(43), false);
-        assert_eq!(TEST_TABLE.contains(u64::MAX), false);
+        assert!(!TEST_TABLE.contains(u64::MIN));
```
brawer: This is now difficult to read. Split the positive and negative test cases into two groups. Still in the same test case, just visually separated in two blocks.

src/tables/u64_set.rs
```diff
@@ -240,11 +240,11 @@ mod tests {
 
         let set = U64Set::create([42, 7, 23, 7].into_iter(), workdir.path(), &out)?;
         assert_eq!(set.len(), 3);
-        assert_eq!(set.contains(0), false);
-        assert_eq!(set.contains(7), true);
-        assert_eq!(set.contains(23), true);
-        assert_eq!(set.contains(42), true);
-        assert_eq!(set.contains(99), false);
+        assert!(!set.contains(0));
```
brawer: Likewise, split the positive and negative test cases into two groups. Still in the same test case, just visually separated.

src/tables/u64_set.rs
```diff
@@ -321,11 +321,11 @@ mod writer {
             assert_eq!(num_written, 3);
 
             let table = U64Set::open(&out)?;
-            assert_eq!(table.contains(4), false);
-            assert_eq!(table.contains(23), true);
-            assert_eq!(table.contains(42), true);
-            assert_eq!(table.contains(7777), true);
-            assert_eq!(table.contains(123), false);
+            assert!(!table.contains(4));
```
brawer: Likewise.

src/tables/u64_set.rs
```diff
@@ -339,10 +339,10 @@ mod writer {
             assert_eq!(num_written, 1);
 
             let table = U64Set::open(&out)?;
-            assert_eq!(table.contains(4), false);
-            assert_eq!(table.contains(8), false);
-            assert_eq!(table.contains(9), true);
-            assert_eq!(table.contains(10), false);
+            assert!(!table.contains(4));
```
brawer, line 347: Likewise.

src/coverage.rs
```diff
@@ -721,12 +722,12 @@ mod writer {
                 );
             }
 
-            assert_eq!(cov.contains_wikidata_item(1), false);
-            assert_eq!(cov.contains_wikidata_item(23), true);
-            assert_eq!(cov.contains_wikidata_item(51), false);
-            assert_eq!(cov.contains_wikidata_item(77), true);
-            assert_eq!(cov.contains_wikidata_item(88), true);
-            assert_eq!(cov.contains_wikidata_item(89), false);
+            assert!(!cov.contains_wikidata_item(1));
```
brawer: Likewise.

src/tables/u64_set.rs
brawer: Done in 4743ce5.

src/tables/u64_set.rs
brawer: Done in 4743ce5.

src/tables/u64_set.rs
brawer: Done in 4743ce5.

src/tables/u64_set.rs
brawer: Done in 4743ce5.

src/coverage.rs
brawer: Done in 4743ce5.

PR #580, opened 2026-08-10
ci: make clippy step gate on warnings and lint tests

brawer:
'cargo clippy --locked' exits 0 even when clippy reports warnings — only
compiler errors fail the step — so this step has effectively been
advisory-only. Without `--all-targets` it also never lints code under
`tests/`.

Verified locally that `cargo clippy --locked --all-targets -- -D warnings`
currently passes clean on this codebase.

Part of #576.


brawer (commented):
LGTM

PR #581, opened 2026-08-10
ci: cap test job at 20 minutes

brawer:
Without `timeout-minutes`, a hang defaults to GitHub's 6-hour cap instead
of failing fast. `test-container.yml` already sets a timeout; `test.yml`
was missing one. 20 minutes is a generous baseline given current build
times; can be tightened later if we get a better sense of typical
duration.

Part of #576.


brawer (commented):
LGTM

PR #582, opened 2026-08-10
ci: remove redundant llvm-tools-preview install step

brawer:
`rust-toolchain.toml` already lists `llvm-tools` in its `components`, so
rustup provisions it automatically the first time cargo/rustc runs in
this checkout (the checkout step already runs before this one). The
explicit `rustup component add llvm-tools-preview` step was a no-op.

This PR's CI run itself is the verification: if `cargo-llvm-cov` can
still generate coverage without the explicit step, the toolchain-file
provisioning is doing its job.

Part of #576.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,8 +35,6 @@ jobs:
       run: |
         sudo apt-get install -y tippecanoe
         tippecanoe --version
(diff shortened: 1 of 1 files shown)


brawer (commented):
LGTM

PR #583, opened 2026-08-10
ci: merge the two apt-get install steps into one

brawer:
protobuf-compiler and tippecanoe were installed in separate steps with
separate `apt-get install` invocations. Merge them into a single step,
and add an explicit `apt-get update` first for robustness against stale
package indexes on the runner image.

Part of #576.


brawer (commented):
LGTM

PR #584, opened 2026-08-10
ci: cache the Cargo registry to speed up dependency fetches

brawer:
sccache caches compilation artifacts but not downloaded crate sources,
so `cargo fetch` re-downloads the full dependency graph (arrow, parquet,
tokio, aws-lc-rs, etc.) on every run. Cache `~/.cargo/registry` and
`~/.cargo/git`, keyed on `Cargo.lock`, to avoid that.

Part of #576.


brawer (commented):
LGTM

PR #585, opened 2026-08-10
Add structured JSON logging to a file in the working directory

brawer:
## Summary
- Add `src/logging.rs`: a skeleton for structured logging. `logging::init(workdir)` wires up the `log` crate's global logger to append one JSON object per line to `<workdir>/pipeline.log` (`timestamp`, `level`, `target`, `message`), instead of the previous `env_logger::init()` to stderr. `RUST_LOG` still controls the level (default `info`).
- At pipeline startup, log an `INFO` record with the binary's version (`CARGO_PKG_VERSION`, formatted like the release tag, e.g. `"v0.6.9"`) and a timestamp.
- In `pipeline::osm::assemble`, replace the `println!` for a relation whose geometry couldn't be built with `log::error!`, so it shows up in the log as e.g. `"could not build geometry for relation/12345"` instead of going to stdout.

This is a first skeleton: per-call structured fields (e.g. a dedicated `relation_id` key rather than embedding it in the message text) aren't wired through `log`'s key-value API yet, and the remaining `print!`/`eprintln!` call sites (`edits.rs`, `tiles.rs`, `upload.rs`) are left for a follow-up.

## Testing
- `cargo test` (full suite, including the `test_pipeline` integration test against the real binary) — all pass.
- `cargo clippy --all-targets -- -D warnings` — clean.
- `cargo fmt --check` — clean.
- Manually ran the pipeline against the test fixtures and confirmed `pipeline.log` is created in the workdir with valid JSON lines, e.g.:
  ```json
  {"level":"INFO","message":"pipeline starting up, version=v0.6.9","target":"osm_diffs::logging","timestamp":"2026-08-10T10:22:00.643612Z"}
  ```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4,6 +4,7 @@ use std::path::PathBuf;
 mod atp;
 mod coverage;
 mod geometry;
+mod logging;
 mod matchers;
 mod pipeline;
 mod places;
--- a/src/logging.rs
+++ b/src/logging.rs
@@ -1,0 +1,100 @@
+//! Structured logging to a file in the working directory.
+//!
+//! [`init`] wires up the `log` crate's global logger (the same facade
+//! already used via `log::info!` etc. elsewhere in this crate) so that
+//! every log record is appended as one JSON object per line to
+//! `<workdir>/pipeline.log`, rather than printed as free-form text. That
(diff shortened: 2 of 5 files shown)


brawer (commented):
LGTM

PR #586, opened 2026-08-10
ci: add a coverage floor to the test run

brawer:
Coverage is reported to Coveralls but nothing previously failed CI if it
regressed. `--fail-under-lines <MIN>` fails the build if total line
coverage drops below `MIN` percent (as opposed to `--fail-uncovered-lines`,
which caps an absolute count of uncovered lines).

Current line coverage is ~91.3% per coveralls.io; set the floor to 88%,
leaving some headroom for noise while still catching real regressions.
Worth ratcheting up over time.

Part of #576.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -49,7 +49,11 @@ jobs:
       run: cargo build --locked
     - name: Run tests
       run: |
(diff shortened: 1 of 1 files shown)


brawer (commented):
LGTM

PR #587, opened 2026-08-10
Log warnings/errors for skipped geometry, replace remaining print!/eprintln!

brawer:
## Summary

Follow-up to #585, which added structured JSON logging to a file in the working directory.

**`pipeline::osm::assemble`** — log when geometry can't be built, instead of silently dropping the feature:
- `assemble_ways`: `log::warn!` when a way's geometry can't be built (degenerate/non-finite coords) -- mirrors the existing relation case.
- `assemble_leaf_relations`: `log::error!` when a leaf relation's geometry can't be built. This had no log statement at all before, even though the identical failure for super-relations already did (added in #585).

**Remaining `print!`/`eprintln!` cleanup** -- everything now funnels into `pipeline.log`:
- `edits.rs`: cache-hit-rate summary → `log::info!`; a permanently disabled `if false { println!(...) }` debug block → unconditional `log::debug!`, so it's actually reachable via `RUST_LOG=debug` instead of being dead code.
- `upload.rs`: missing `S3_ENDPOINT` (upload skipped) → `log::warn!`.
- `tiles.rs`: tippecanoe's non-progress stderr lines and dropped empty layers → `log::info!`.

`rg "print!|println!|eprintln!|eprint!" src` now returns nothing.

## Testing
- `cargo test` (full suite, including `test_pipeline` integration test) — all pass.
- `cargo clippy --all-targets -- -D warnings` — clean.
- `cargo fmt --check` — clean.

### DIFF
--- a/src/pipeline/edits.rs
+++ b/src/pipeline/edits.rs
@@ -79,8 +79,8 @@ pub fn suggest_edits(
     progress_bar.finish_with_message(format!("ATP features → {} suggested OSM edits", num_edits));
 
     let cache_stats = osm.cache_stats();
-    println!(
-        "  cache hits: {} misses: {} hit rate: {:.1}%",
+    log::info!(
+        "cache hits: {} misses: {} hit rate: {:.1}%",
         cache_stats.hits,
         cache_stats.misses,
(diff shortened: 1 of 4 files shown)


brawer (commented):
LGTM

PR #590, opened 2026-08-10
Add release-process and supply-chain-security documentation

brawer:
New `docs/` directory, matching the existing "README per directory" convention already used for `scripts/` and `scripts/sbom/`.

- **`docs/README.md`**: index.
- **`docs/releasing.md`**: the practical how-to for cutting a release -- quick start, how to choose the version number (SemVer driven by output-schema compatibility, not code churn), what happens automatically step by step, how to verify a release actually came out right, rules (always use the script, releases are immutable, who's allowed to release), what to do if something goes wrong at each stage, where things live, and known gaps that are deliberately out of scope for now (no production deployment yet; links to #555/#556/#588/#589 for the rest).
- **`docs/supply-chain-security.md`**: the concepts behind all of this -- Bill of Materials (framed generally, not just "software," since a future BOM-for-output-data is plausible per #588), SBOM, CBOM, CycloneDX, build provenance/attestations (including why per-architecture images get their own SBOM but the manifest list only gets provenance, linking to the research on #589), immutable releases, and a candid section on whether any of this is overkill for a project like this.
- **`scripts/verify-release.sh`**: added per review feedback -- automates the SBOM/build-provenance attestation checks that were originally described in `docs/releasing.md` as manual copy-paste commands. Wired into `cut-release.sh` as its final step, and also standalone-invokable (e.g. if `cut-release.sh` gets interrupted, or to audit an older release later).

Also trims `scripts/sbom/README.md` down to implementation only (the `.jq` pipeline, dev-mode placeholder behavior), removing the conceptual SBOM/CBOM explanation now that it lives in `docs/supply-chain-security.md`, and adds a one-line pointer from the top-level `README.md` to `docs/`.

## Testing
- Verified every cross-referenced file path and every markdown anchor link resolves to a real file/heading.
- `scripts/verify-release.sh`: exercised all three code paths against real GitHub state, not just read through:
  - **Success path**: ran it against the real, already-published `v0.6.9` release -- correctly found the `release.yml` run, confirmed it succeeded, downloaded both per-architecture digests, and confirmed both a build-provenance (`https://slsa.dev/provenance/v1`) and an SBOM (`https://cyclonedx.org/bom`) attestation exist for each.
  - **Not-found/wait path**: confirmed the `gh run list` query correctly returns empty for a tag with no matching run (`v99.99.99`), which is what makes the script's poll loop keep waiting rather than misfiring.
  - **Failure path**: ran it against `v0.6.7`, a real historically-failed `release.yml` run (from before #558-#561 fixed the underlying issues) -- correctly detected `conclusion=failure` and exited non-zero with a link to the failed run, instead of hanging or reporting success.
  - Also `sh -n` syntax-checked both `cut-release.sh` and `verify-release.sh` after wiring them together.

### DIFF
--- a/docs/releasing.md
+++ b/docs/releasing.md
@@ -14,13 +14,23 @@ This document is the practical how-to for cutting a release of
 Run it from a clean, up-to-date `main` checkout. It handles everything
 else: bumping the version, opening and merging the PR for that, tagging,
 and publishing the release. See “What happens automatically” below for
-the full sequence.
+the full sequence — but read “Choosing the version number” first, it’s
+the one part of this that actually needs a careful decision, not just
+running a command.
 
 ## Choosing the version number
 
 This is the one genuine judgment call in the whole process; everything
 else is mechanical. We follow [SemVer](https://semver.org/), driven by
-the pipeline’s **output schema** — not by how much code changed:
+the pipeline’s **output schema** — not by how much code changed.
+
+This is a different question than what [SemVer](https://semver.org/)
+usually answers. Programmers normally think of SemVer in terms of API
+compatibility for code that *links against* a library. Nobody links
+against `osm-diffs`:
+downstream clients only ever consume the *data* it produces. So the
+question to ask isn’t “did the code change in a breaking way,” it’s “does
+this change what a client reading our output has to handle differently”:
 
 - **major** — the output schema changed in a way that breaks existing
   clients (a field was removed or renamed, a type changed, ...)
@@ -63,7 +73,8 @@ Once you run `cut-release.sh vX.Y.Z`:
      (a server-side safety net — this should never fail if you used the
      script, since the script only ever tags a version it just set).
    - `build` (once per architecture, amd64 and arm64): builds the
-     container via `Containerfile`, which also generates the SBOM (see
+     container via [`Containerfile`](../Containerfile), which also
+     generates the SBOM (see
      [`scripts/sbom/README.md`](../scripts/sbom/README.md)) and pushes
      each architecture’s image to `ghcr.io` by digest.
    - `manifest`: combines both architectures into a multi-arch manifest,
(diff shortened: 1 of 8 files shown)


docs/releasing.md
```diff
@@ -0,0 +1,168 @@
...
+Run it from a clean, up-to-date `main` checkout. It handles everything
+else: bumping the version, opening and merging the PR for that, tagging,
+and publishing the release. See “What happens automatically” below for
+the full sequence.
+
+## Choosing the version number
+
+This is the one genuine judgment call in the whole process; everything
+else is mechanical. We follow [SemVer](https://semver.org/), driven by
+the pipeline’s **output schema** — not by how much code changed:
```
brawer: Very briefly, explain why: Downstream clients consume _data_ from our pipeline, they don't link to any code. IMHO this warrants a (very brief) explanation, since it may not be obvious to programmers who typically think about API versions, not data dependencies.

docs/releasing.md
```diff
@@ -0,0 +1,168 @@
...
+
+This document is the practical how-to for cutting a release of
+`osm-diffs`. For the concepts behind *why* the process looks like this
+(SBOM, attestations, immutable releases, ...), see
+[`supply-chain-security.md`](supply-chain-security.md).
+
+## Quick start
+
+```sh
+./scripts/cut-release.sh vX.Y.Z
```
brawer, line 11: Maybe add a comment, linking to “Choosing the version number”? Quick start is great, but I'm afraid people won't read any further, and choosing the version number is the only thing they really have to do carefully.

docs/releasing.md
```diff
@@ -0,0 +1,168 @@
...
+   immutable from this point on (see
+   [`supply-chain-security.md`](supply-chain-security.md#immutable-releases)).
+5. **That tag push triggers
+   [`.github/workflows/release.yml`](../.github/workflows/release.yml)**,
+   entirely independently of the script:
+   - `verify-version`: re-checks the tag matches `Cargo.toml`’s version
+     (a server-side safety net — this should never fail if you used the
+     script, since the script only ever tags a version it just set).
+   - `build` (once per architecture, amd64 and arm64): builds the
+     container via `Containerfile`, which also generates the SBOM (see
```
brawer: Link to Containerfile.

docs/releasing.md
```diff
@@ -0,0 +1,168 @@
...
+   - `manifest`: combines both architectures into a multi-arch manifest,
+     tagged both `vX.Y.Z` and `latest`.
+   - `attest`: publishes signed SBOM and build-provenance attestations
+     for both per-architecture images, plus a build-provenance
+     attestation for the manifest list.
+
+Steps 1-4 usually take under 10 minutes; step 5 (the actual container
+build) takes roughly another 15-20 minutes.
+
+## Verifying a release
```
brawer, line 94: Instead of describing manual steps, can't we automate this verification? Perhaps a separate script in `scripts`, invoked from `cut-release.sh`. It's totally okay of cut-release takes 30mins to complete, but if the machine crashes or whatver, then people could still call the verification script.

docs/supply-chain-security.md
```diff
@@ -0,0 +1,106 @@
...
+
+Our **S**oftware **B**ill of **M**aterials describes the container image
+we publish: the Rust dependency graph of the `osm-diffs` binary, the
+statically linked `tippecanoe` binary and the libraries baked into it,
+build-environment details (compiler and OS versions), and licenses.
+
+Alongside it, our SBOM includes a small **C**ryptographic **B**ill of
+**M**aterials (CBOM): the same idea, but for cryptography instead of
+dependencies — which TLS version, cipher suites, and crypto backend the
+binary uses. This matters because that’s exactly the kind of thing that
```
brawer: After "binary uses", add: ", for example when uploading its output to S3-compatible storage."

docs/supply-chain-security.md
```diff
@@ -0,0 +1,106 @@
...
+
+Attestations prove a release was built correctly *once*. Immutability
+protects that guarantee over time: once a release is published, GitHub
+locks its git tag to that exact commit — it can’t be moved, and it can’t
+be deleted while the release exists. Even if the release is deleted
+outright, its tag name can never be reused for a new release. This closes
+off a whole class of supply-chain attack where a previously-trusted,
+already-verified tag gets quietly repointed at something else later.
+
+## Why all of this, together
```
brawer, line 120: Add something along these lines: Isn’t this total over-engineering for a pipeline whose code is open-source, that processes freely available data, and that doesn't handle anything secret. So, supply chain security isn't as important as is for many other projects. However, once set up, it's entirely automated, so why not doing it properly..

docs/releasing.md
brawer: Added a paragraph explaining this: clients consume our *data*, they don't link against our *code*, so the usual "did the API change" framing doesn't apply -- the question is whether the output schema changed.

docs/releasing.md
brawer: Added a pointer from Quick start straight to "Choosing the version number".

docs/releasing.md
brawer: Linked.

docs/releasing.md
brawer: Agreed, and done: added scripts/verify-release.sh, wired into cut-release.sh as its final step (with a note that it's safe to Ctrl-C and re-run standalone later if the machine dies mid-wait). "Verifying a release" now documents that instead of manual steps. Tested against the real v0.6.9 release, a nonexistent tag, and a real historically-failed release to cover the success/wait/failure paths.

docs/supply-chain-security.md
brawer: Added.

docs/supply-chain-security.md
brawer: Added a new "Is this overkill for a project like this?" section addressing this directly.

brawer (commented):
LGTM

PR #591, opened 2026-08-10
Move CODE_OF_CONDUCT.md and SECURITY.md into docs/

brawer:
Follow-up from #590 -- grouping the rest of the project's docs together now that \`docs/\` exists.

GitHub recognizes community health files (README, CODE_OF_CONDUCT, SECURITY, ...) in the repo root, \`.github/\`, or \`docs/\` equally -- verified against GitHub's own docs, not assumed -- so this doesn't lose any special UI treatment (Community tab, Security tab).

\`README.md\` stays at root: it already collides with \`docs/README.md\` (the \`docs/\` directory's own index), and root is the first place anyone looks anyway, so there's nothing to gain by moving it.

Also fixes a stale link while touching \`SECURITY.md\`: it pointed at \`https://github.com/diffed-places/diffed-places-pipeline/security\`, an old repo name from before this project was renamed. GitHub redirects it (confirmed via \`curl -I\`), so it wasn't actually broken, but there's no reason to rely on a redirect when the fix is one line.

\`docs/README.md\` updated to link both.

## Testing
Checked baseline via \`gh api repos/alltheplaces/osm-diffs/community/profile\` before merging (confirms \`CODE_OF_CONDUCT.md\` is currently found at root, and that GitHub already recognizes this repo's \`docs/\` folder as its "documentation" location). Will re-check the same endpoint plus the actual Security tab after merging to confirm both files are still picked up correctly from their new location.


brawer (commented):
LGTM

PR #592, opened 2026-08-10
Rename docs/*.md to consistent uppercase filenames

brawer:
Follow-up from #591.

\`docs/releasing.md\` -> \`docs/RELEASING.md\`
\`docs/supply-chain-security.md\` -> \`docs/SUPPLY_CHAIN_SECURITY.md\`

Matches \`CODE_OF_CONDUCT.md\` and \`SECURITY.md\`, which need their exact uppercase names for GitHub to recognize them as community health files -- rather than have those two be uppercase by necessity and the other two lowercase by accident, make all of \`docs/\` consistent. \`README.md\` is unchanged (already uppercase, and is itself one of the GitHub-recognized names).

Updated every cross-reference (both the link text and the path) across \`docs/README.md\`, \`docs/RELEASING.md\`, \`docs/SUPPLY_CHAIN_SECURITY.md\`, \`scripts/README.md\`, \`scripts/verify-release.sh\`, and \`scripts/sbom/README.md\`.

## Testing
Grepped the whole repo for both old filenames after the change -- no references remain anywhere. Confirmed the anchored links (\`#immutable-releases\`, \`#build-provenance-and-attestations\`) still resolve, since the headings they point at didn't change, only the filename. \`sh -n\` on \`verify-release.sh\` (only its header comment changed) still passes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #593, opened 2026-08-10
Fix flaky logging test: search for our records instead of assuming exact position/count

brawer:
Found while investigating why CI was failing on #592 -- a docs-only rename that couldn't plausibly have caused a Rust test failure: https://github.com/alltheplaces/osm-diffs/actions/runs/31425095611/job/93574785212

\`test_init_writes_structured_startup_record\` calls \`init()\`, which installs the process-wide \`log::set_logger\` -- per the existing comment, that can only happen once per process, so it's the only test in the crate that calls it. But that logger then stays live for the rest of the test binary's process, and \`cargo test\` runs tests in parallel within one process: any other test that logs anything afterwards -- directly, or via a dependency such as \`ext_sort\` -- gets appended to this same file, since there's only one global logger for the whole process at that point.

The test asserted exactly 2 lines at fixed positions, which is racy: it depends on whether some other concurrently-running test happens to log something in the window between \`init()\` and the final read. That's exactly what happened in the linked run -- an \`ext_sort\` thread-pool-init log line interleaved in, and the test saw 3 lines instead of 2.

## Fix
Search the parsed records for our specific startup record (by target + level) and our specific warn record (by target + level), rather than assuming line count or position. This tolerates any number of interleaved lines from other tests while still verifying the same thing: that \`init()\` produces a correct structured startup record, and that our own \`log::warn!\` call is captured correctly.

## Testing
- \`cargo build\`, \`cargo fmt --check\`, \`cargo clippy --locked\`: all clean.
- Ran the full \`cargo test --lib\` suite 30 times in a loop (not just this test in isolation, which wouldn't exercise the interleaving at all) -- no failures.
- The fix is correct by construction regardless of whether a given run happens to reproduce the interleaving, since it no longer assumes anything about what else might be in the file.

### DIFF
--- a/src/logging.rs
+++ b/src/logging.rs
@@ -69,6 +69,13 @@ mod tests {
     // `init()` sets the process-wide logger, which -- like `log::set_logger`
     // in general -- can only happen once per process; keep this the only
     // test in the crate that calls it.
+    //
+    // That logger then stays live for the rest of the test binary's
+    // process. `cargo test` runs tests in parallel within one process, so
+    // any other test that logs anything afterwards -- directly, or via a
+    // dependency such as ext_sort -- gets appended to this very same file.
+    // Search for our own two records by content instead of assuming we're
+    // the only lines in the file, or that they're at fixed positions.
     #[test]
     fn test_init_writes_structured_startup_record() -> Result<()> {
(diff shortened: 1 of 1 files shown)


brawer (commented):
LGTM

PR #594, opened 2026-08-10
docs/SUPPLY_CHAIN_SECURITY.md: move "overkill?" section to the end, trim BOM section

brawer:
Small editorial follow-up from #590/#591.

- Moves "Is this overkill for a project like this?" to after "Why all of this, together" -- reads better as the closing thought than as a mid-document detour.
- Drops the "our SBOM already has data components too" aside from the Bill of Materials section for brevity.


brawer (commented):
LGTM

PR #595, opened 2026-08-11
docs/SUPPLY_CHAIN_SECURITY.md: add Minimal containers section, name SLSA Build L3

brawer:
Adds concepts the docs didn't cover yet, all visible in the code but never explained -- `docs/SUPPLY_CHAIN_SECURITY.md` now reads roughly in this order:

- **Containers** / **Minimal containers** / **Multi-architecture containers** -- new, placed at the front (before "Bill of Materials"), since later sections already presuppose the reader knows what a container is. Covers: what a container is generally (vs. a VM); that we publish an OCI image to a container registry, built from `Containerfile`; why the shipped image is `FROM scratch` with nothing but two statically-linked (musl, not glibc) binaries, no shell, no package manager, running unprivileged -- and why that matters even with code execution; and why/how the image is published for both `amd64` and `arm64` (local debugging on Apple Silicon without needing an old Intel machine; ARM's growing datacenter footprint; mechanically just an OCI image index pointing at per-platform images).
- **SLSA Build Level 3**, named explicitly in "Build provenance and attestations": the one thing that actually earns it is the `attest` job running separately from `build`/`manifest`, so signing credentials are never exposed to the steps that compile arbitrary code.
- **"Is this overkill for a project like this?"**, moved to the very end as the closing thought -- addresses the threat-model question honestly (modest, for an open-source project handling only public data) and the "why bother anyway" question (one-time, fully automated cost; also a bet on where the industry is heading, citing the 2024 xz-utils backdoor).

Every technical claim and background link was checked directly, not asserted from memory -- including several corrections made during review: swapped a misleading Distroless link (we use `scratch`, not any distroless base image) for a verified attack-surface-reduction source after the first replacement also turned out to be the wrong kind of page; corrected the musl-vs-glibc framing after checking musl.libc.org's own design rationale, then dropped the static/dynamic-linking angle entirely per feedback; and deliberately avoided both "Docker" (we use OCI images, built with podman) and "Linux containers" (our own image, built `FROM scratch`, has no Linux userspace inside it) as terms throughout.

Kept every addition short per the "nobody reads long docs" goal -- a few sentences per concept, not new essays.

### DIFF
--- a/docs/SUPPLY_CHAIN_SECURITY.md
+++ b/docs/SUPPLY_CHAIN_SECURITY.md
@@ -40,6 +40,22 @@ changed.
 Both are generated from real, current build state, not maintained by
 hand — see [`scripts/sbom/README.md`](../scripts/sbom/README.md) for how.
 
+## Minimal containers
+
+`Containerfile` builds in two stages, and only the second one ships:
+`FROM scratch`, containing nothing but the `osm-diffs` and `tippecanoe`
+binaries. The entire build toolchain (Rust, a C compiler, `apk`, `git`,
+...) stays behind in the discarded first stage. Both binaries are
+statically linked against musl (Alpine’s C library) rather than
+dynamically against glibc, so there isn’t even a shared C library in the
+final image — let alone a shell or a package manager. The process also
+runs as an unprivileged user, not `root`. If someone found a way to run
+arbitrary code inside this container, there’s nothing there to run: no
+`/bin/sh`, nothing to fetch a second-stage payload with, and no root
+privileges to do more damage with even so. This is the same idea behind
+[Distroless](https://github.com/GoogleContainerTools/distroless) images,
+not something specific to `osm-diffs`.
+
 ## CycloneDX
 
 We publish our SBOM (and CBOM) in [CycloneDX](https://cyclonedx.org/)
@@ -60,6 +76,14 @@ build provenance attestation records: a cryptographically signed
 statement, tied to the exact artifact digest, that says “this was built
 by workflow W, from commit C, on GitHub-hosted infrastructure, triggered
 by event E.”
+
+This specifically targets
+[SLSA Build Level 3](https://slsa.dev/spec/v1.2/build-track-basics#build-l3):
+provenance that isn’t just signed, but generated somewhere the build
+itself can’t influence or forge. That’s why `release.yml`’s `attest` job
+runs as a separate job from `build`/`manifest` — the credentials used to
+sign the provenance are never exposed to the steps that actually compile
+arbitrary code, which is the part an attacker is more likely to reach.
 
 We use GitHub’s native
 [artifact attestations](https://docs.github.com/en/actions/security-guides/using-artifact-attestations-to-establish-provenance-for-builds)
--- a/docs/SUPPLY_CHAIN_SECURITY.md
+++ b/docs/SUPPLY_CHAIN_SECURITY.md
@@ -15,10 +15,18 @@ to run — binaries, libraries, configuration — into a single, portable
 unit that behaves the same way regardless of the machine underneath it.
 It’s much lighter weight than a full virtual machine, since it shares
 the host machine’s kernel instead of running its own, while still
-keeping what’s inside isolated from everything else on that host. We
-publish `osm-diffs` as an [OCI](https://opencontainers.org/) image — the
-open, vendor-neutral standard most container tooling implements today —
-to GitHub’s container registry.
+keeping what’s inside isolated from everything else on that host.
+
+We publish `osm-diffs` as an [OCI](https://opencontainers.org/) image —
+the open, vendor-neutral standard most container tooling implements
+today — to a
+[container registry](https://www.redhat.com/en/topics/cloud-native-apps/what-is-a-container-registry)
+(GitHub’s, specifically): a place to store and distribute container
+images, the same way a package registry does for libraries.
+[`Containerfile`](../Containerfile) is the
+[recipe that builds our image](https://github.com/containers/container-libs/blob/main/common/docs/Containerfile.5.md):
+a plain-text file listing the steps to assemble it, one instruction per
+line.
 
 ## Minimal containers
 
@@ -37,6 +45,24 @@ arbitrary code inside this container, there’s nothing there to run: no
 privileges to do more damage with even so — see
 [this explanation of container attack-surface reduction](https://www.minimus.io/post/container-image-attack-surface-reduction)
 for more background on why that matters.
+
+## Multi-architecture containers
+
+Our container image is published for both `amd64` and `arm64`. The main
+reason: this is meant to run as a weekly batch job in a datacenter, but
+if something goes wrong, a developer should be able to just pull the
+image and run it locally to debug — including on an Apple Silicon Mac,
+without hunting down an old Intel machine first. ARM is also more
+power-efficient than x86 and gaining ground in datacenter infrastructure
+for that reason; publishing both architectures now makes an eventual
+production move easier, if that ever happens.
+
+Mechanically, a multi-architecture image is just an
+[image index](https://github.com/opencontainers/image-spec/blob/main/image-index.md)
+(also called a manifest list): a small document that points to one
+architecture-specific image per platform. Pulling the image by its tag
+automatically fetches the right one for your machine — you never have to
+pick `amd64` or `arm64` yourself.
 
 ## Bill of Materials (BOM)
 
--- a/docs/SUPPLY_CHAIN_SECURITY.md
+++ b/docs/SUPPLY_CHAIN_SECURITY.md
@@ -17,7 +17,8 @@ It’s much lighter weight than a full virtual machine, since it shares
 the host machine’s kernel instead of running its own, while still
 keeping what’s inside isolated from everything else on that host.
 
-We publish `osm-diffs` as an [OCI](https://opencontainers.org/) image —
+We publish `osm-diffs` as an
+[OCI image](https://opencontainers.org/) —
 the open, vendor-neutral standard most container tooling implements
 today — to a
 [container registry](https://www.redhat.com/en/topics/cloud-native-apps/what-is-a-container-registry)
--- a/docs/SUPPLY_CHAIN_SECURITY.md
+++ b/docs/SUPPLY_CHAIN_SECURITY.md
@@ -56,9 +56,9 @@ shell or a package manager. The process also
 runs as an unprivileged user, not `root`. If someone found a way to run
 arbitrary code inside this container, there’s nothing there to run: no
 `/bin/sh`, nothing to fetch a second-stage payload with, and no root
-privileges to do more damage with even so. This is the same idea behind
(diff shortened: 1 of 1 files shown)


docs/SUPPLY_CHAIN_SECURITY.md
```diff
@@ -61,6 +77,14 @@ statement, tied to the exact artifact digest, that says “this was built
 by workflow W, from commit C, on GitHub-hosted infrastructure, triggered
 by event E.”
 
+This specifically targets
```
brawer: Change to "We target"

docs/SUPPLY_CHAIN_SECURITY.md
```diff
@@ -40,6 +40,22 @@ changed.
 Both are generated from real, current build state, not maintained by
 hand — see [`scripts/sbom/README.md`](../scripts/sbom/README.md) for how.
 
+## Minimal containers
+
+`Containerfile` builds in two stages, and only the second one ships:
+`FROM scratch`, containing nothing but the `osm-diffs` and `tippecanoe`
+binaries. The entire build toolchain (Rust, a C compiler, `apk`, `git`,
+...) stays behind in the discarded first stage. Both binaries are
+statically linked against musl (Alpine’s C library) rather than
```
brawer: Make musl a link to https://en.wikipedia.org/wiki/Musl

docs/SUPPLY_CHAIN_SECURITY.md
```diff
@@ -40,6 +40,22 @@ changed.
...
+dynamically against glibc, so there isn’t even a shared C library in the
```
brawer: It's not really about static vs dynamic linking. See https://tuxcare.com/blog/musl-vs-glibc/ — actually, this might be a good link to insert into the doc, so readers can learn more about musl. Maybe there's also better sources.

docs/SUPPLY_CHAIN_SECURITY.md
```diff
@@ -40,6 +40,22 @@ changed.
...
+final image — let alone a shell or a package manager. The process also
+runs as an unprivileged user, not `root`. If someone found a way to run
+arbitrary code inside this container, there’s nothing there to run: no
+`/bin/sh`, nothing to fetch a second-stage payload with, and no root
+privileges to do more damage with even so. This is the same idea behind
+[Distroless](https://github.com/GoogleContainerTools/distroless) images,
+not something specific to `osm-diffs`.
```
brawer: Nothing of what we're doing is remotely original; everything described is simply good practice. Actually, we should say this somewhere in the doc. Perhaps at the very beginning, as an end of the intro paragraph on line 7?

docs/SUPPLY_CHAIN_SECURITY.md
brawer: Done.

docs/SUPPLY_CHAIN_SECURITY.md
brawer: Done.

docs/SUPPLY_CHAIN_SECURITY.md
brawer: Checked musl.libc.org's own design page and used that instead: musl's real point is simplicity/size/standards-compliance, static-linking-friendliness is a consequence of that, not the defining trait. Then per your follow-up, dropped the static-vs-dynamic comparison to glibc entirely rather than just reframing it.

docs/SUPPLY_CHAIN_SECURITY.md
brawer: Added at the end of the intro paragraph, as suggested. Left out "widely used" per your follow-up -- not confident that's actually true industry-wide today.

docs/SUPPLY_CHAIN_SECURITY.md
brawer: Having a background link here is a very good idea, but the specific link is misleading. We don't actually use any of Google's "distroless" base images; rather, we have no base image at all: Our second-stage container is built from scratch, not `gcr.io/distroless/*`. Instead of pointing to the Distroless project, can you link to some background page that explains the concept and benefits of _scratch_ containers?

brawer: Omit (for brevity). The intro already says we're not doing anything novel.

brawer: Good catch. Swapped the link -- and my first replacement (Docker's own base-images docs) turned out to have the same problem you're about to point out, so see the follow-up commit too.

brawer: Removed.

docs/SUPPLY_CHAIN_SECURITY.md
```diff
@@ -40,6 +42,24 @@ changed.
 Both are generated from real, current build state, not maintained by
 hand — see [`scripts/sbom/README.md`](../scripts/sbom/README.md) for how.
 
+## Minimal containers
```
brawer: Very nice, thank you. One thing, though: Before we describe _Minimal containers_, we should explain what Containers are. A lot of developers don't know this (or have only a vague idea, along the lines of 'this hot thing everyone uses'). Add a brief section titled "Containers" that explains, very briefly, the idea and benefits. However, please don't use "docker" in the text. We're actually using OCI containers, not the proprietary original docker format, and also our build chain uses podman instead of docker because we're an open-source project so we don't use proprietary tools (but no need to go talk about this here, let's not bash docker). Some people use the term "Linux containers", but that wouldn't be a good phrase to use because we're building our containers from scratch, so there's no Linux userspace.

For the overall reading flow, should we introduce "Containers" and "Minimal Containers" before we talk about BOMs? Leaving this up to you.

brawer: Added a brief "Containers" section, avoiding both "Docker" and "Linux containers" as terms. Moved it and "Minimal containers" to the front, before BOM -- you're right that "SBOM and CBOM" already presupposed the reader knew what a container was, so the dependency order points that way.

docs/SUPPLY_CHAIN_SECURITY.md
```diff
@@ -4,7 +4,39 @@ This document explains the concepts behind the supply-chain security
...
+
+A container packages an application together with everything it needs
+to run — binaries, libraries, configuration — into a single, portable
+unit that behaves the same way regardless of the machine underneath it.
+It’s much lighter weight than a full virtual machine, since it shares
+the host machine’s kernel instead of running its own, while still
+keeping what’s inside isolated from everything else on that host. We
+publish `osm-diffs` as an [OCI](https://opencontainers.org/) image — the
+open, vendor-neutral standard most container tooling implements today —
+to GitHub’s container registry.
```
brawer: Should we explain what a container registry is? Perhaps just make "container registry" or "GitHub container registry" a link. Maybe to https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry although that page is about technicalities and pricing, it would be better to link to a page that introduces the concept of container registry at a high level.

docs/SUPPLY_CHAIN_SECURITY.md
```diff
@@ -4,7 +4,39 @@ This document explains the concepts behind the supply-chain security
...
+
+## Minimal containers
+
+`Containerfile` builds in two stages, and only the second one ships:
```
brawer, line 34: Many people don't know what a Containerfile is. Explain the purpose. Maybe just one sentence, perhaps at the end of the previous section after having introduced Containers, with a link to the spec or some nice tutorial; don't go into depth. (Don't use the term Dockerfile, although it's the same thing).

docs/SUPPLY_CHAIN_SECURITY.md
```diff
@@ -4,7 +4,39 @@ This document explains the concepts behind the supply-chain security
 practices used in this project’s release process (see
 [`RELEASING.md`](RELEASING.md) for the actual, repo-specific steps). It’s
 written for someone who can code but hasn’t necessarily done release
-engineering before.
+engineering before. None of it is original to `osm-diffs` — everything
+described here is standard practice; this document just explains it in
+one place.
+
+## Containers
```
brawer, line 11: Somewhere, we also need to introduce multi-architecture containers, as we talk about this later down in the document (when discussing that each arch gets their own SBOM). Maybe a new section titled "Multi-architecture containers", after "Minimal containers", before introducting BOMs? Start with the _why_: Many developers have ARM machines (Apple silicon); although our container is mainly intended for running as a weekly cronjob in a datacenter, in case of a problem, it's good if developers can simply pull the container to their machine and run it locally, without being forced to find an old Intel laptop. Also, since ARM is less power-hungry than Intel, it's getting some traction for datacenter infrastructure; by doing multi-arch, we can ease an eventual transition in production. — After the _why_, introduce the _how_: Explain that a multi-arch container is just a manifest that points to arch-specific images. Please keep it as brief as possible, since nobody reads long docs, but link to background where it makes sense.

docs/SUPPLY_CHAIN_SECURITY.md
brawer: Linked to Red Hat's conceptual overview instead -- clear, high-level, avoids the technical/pricing framing of the GitHub docs page you found.

docs/SUPPLY_CHAIN_SECURITY.md
brawer: Added one sentence + a link to Podman's own Containerfile(5) spec page (good "automates the steps of creating a container image, similar to a Makefile" framing, and it names Podman/Buildah/Docker together rather than being Docker-branded). Avoided the term "Dockerfile" in our own prose.

docs/SUPPLY_CHAIN_SECURITY.md
brawer: Added, placed exactly where you suggested. Led with the why (local debugging on Apple Silicon, ARM's growing datacenter footprint) before the how (image index/manifest list). Checked the OCI image-index spec directly -- it's accurate but too dry to carry the why on its own, so wrote that ourselves and linked the spec for the formal definition.

docs/SUPPLY_CHAIN_SECURITY.md
```diff
@@ -4,7 +4,65 @@ This document explains the concepts behind the supply-chain security
...
+
+A container packages an application together with everything it needs
+to run — binaries, libraries, configuration — into a single, portable
+unit that behaves the same way regardless of the machine underneath it.
+It’s much lighter weight than a full virtual machine, since it shares
+the host machine’s kernel instead of running its own, while still
+keeping what’s inside isolated from everything else on that host.
+
+We publish `osm-diffs` as an [OCI](https://opencontainers.org/) image —
```
brawer: Make "OCI image" the link anchor (make "image" clickable too). Otherwise, the link is difficult to spot in a long sea of text, "OCI" is just three letters.

brawer: Done.

brawer (commented):
LGTM

PR #596, opened 2026-08-11
Tighten SUPPLY_CHAIN_SECURITY.md prose, add a supply-chain definition

brawer:
Fresh read-through as the target reader (can code, no release-engineering or industry background): found one real comprehension gap and a lot of length to cut without losing any concept.

- The doc never defined "software supply chain" before diving into specific practices -- added one sentence to the intro.
- "Sigstore" was a completely bare proper-noun link; added a three-word gloss.
- Tightened prose throughout: cut redundant clauses, collapsed triple-stacked examples into pairs where a third example added length without adding information, removed hedge words that didn't carry meaning.

**1449 -> 1181 words (18% shorter).**

## Testing
- Confirmed every concept from the previous version is still present (went concept-by-concept, not just word-count-driven).
- No straight quotes/dashes crept back in; "docker" and "linux container" still don't appear anywhere (per the constraints from #595's review).
- Both anchored cross-references from `RELEASING.md` (`#immutable-releases`, `#build-provenance-and-attestations`) still resolve, since heading text and order are unchanged.

### DIFF
--- a/docs/SUPPLY_CHAIN_SECURITY.md
+++ b/docs/SUPPLY_CHAIN_SECURITY.md
@@ -78,13 +78,11 @@ Our SBOM describes the container image we publish: the Rust dependency
 graph of the `osm-diffs` binary, the statically linked `tippecanoe`
 binary and its libraries, build-environment details, and licenses.
 
-Alongside it, our SBOM includes a small Cryptographic Bill of Materials
+Alongside it, our SBOM includes a small
+[Cryptographic Bill of Materials](https://en.wikipedia.org/wiki/Cryptographic_bill_of_materials)
 (CBOM): the same idea, but for cryptography — which TLS version, cipher
 suites, and crypto backend the binary uses, for example when uploading
-its output to S3-compatible storage. This matters because it’s exactly
-the kind of thing that can go stale silently: a routine dependency
-update wouldn’t otherwise tell you the set of cipher suites your binary
-supports has changed.
+its output to S3-compatible storage.
 
 Both are generated from real, current build state, not maintained by
 hand — see [`scripts/sbom/README.md`](../scripts/sbom/README.md) for how.
@@ -164,8 +162,8 @@ But it’s a one-time cost: once set up, it runs fully automated, adding
 no effort to any future release. Cheap insurance, not a real trade-off
 against other work.
 
-It’s also a bet on where the industry is heading, not just where it
-stands today: supply-chain attacks aren’t hypothetical anymore (the
+It’s also about where things are heading, not just where they stand
+today: supply-chain attacks aren’t hypothetical anymore (the
 [xz-utils backdoor](https://en.wikipedia.org/wiki/XZ_Utils_backdoor)
 being the starkest recent example), and what this document describes
 is becoming a baseline expectation, not something only
--- a/docs/SUPPLY_CHAIN_SECURITY.md
+++ b/docs/SUPPLY_CHAIN_SECURITY.md
@@ -1,100 +1,90 @@
 # Supply-chain security: concepts
 
-This document explains the concepts behind the supply-chain security
-practices used in this project’s release process (see
-[`RELEASING.md`](RELEASING.md) for the actual, repo-specific steps). It’s
-written for someone who can code but hasn’t necessarily done release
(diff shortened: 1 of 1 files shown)


docs/SUPPLY_CHAIN_SECURITY.md
```diff
@@ -1,192 +1,172 @@
...
 
-Alongside it, our SBOM includes a small **C**ryptographic **B**ill of
-**M**aterials (CBOM): the same idea, but for cryptography instead of
-dependencies — which TLS version, cipher suites, and crypto backend the
-binary uses, for example when uploading its output to S3-compatible
-storage. This matters because that’s exactly the kind of thing that
-can go stale silently: nothing about a routine dependency update would
-otherwise tell you that the set of cipher suites your binary supports has
-changed.
+Alongside it, our SBOM includes a small Cryptographic Bill of Materials
```
brawer: Can we shorten this further? The CBOM isn't that important for us. Maybe skip "This matters ... has changed", and just make "Cryptographic Bill of Materials (CBOM)" a link to some site that explains it. Ideally a page that focuses on the why, not the how. Maybe just condense this whole paragraph to introducing CBOM as same idea, just for crypto, for example which TLS version or cipher suites the binary uses when talking to other machines.

docs/SUPPLY_CHAIN_SECURITY.md
```diff
@@ -1,192 +1,172 @@
...
+Fair question. `osm-diffs` is open source, processes only public data,
+and handles no secrets — our actual threat model is modest compared to
+a proprietary service handling user data or credentials. So no, this
+isn’t strictly *necessary* the way it would be for many other projects.
+
+But it’s a one-time cost: once set up, it runs fully automated, adding
+no effort to any future release. Cheap insurance, not a real trade-off
+against other work.
+
+It’s also a bet on where the industry is heading, not just where it
```
brawer: Ugh, "bet" sounds like the voice of an alpha male, please find a less competitive-sounding term. Prediction? Assumption? Description? Or, just say that it's the direction where things are heading, that would be shorter too. (Btw, "industry" is a bit weird for an open-source project; while not alpha-male-y, it sounds a little corporate). Well, you are the better wordsmith than me; I'm curious what you'll come up with.

docs/SUPPLY_CHAIN_SECURITY.md
brawer: Done — dropped the staleness sentence and linked "Cryptographic Bill of Materials" to https://en.wikipedia.org/wiki/Cryptographic_bill_of_materials instead (checked it leads with the why — risk visibility, crypto-agility — not just the CycloneDX field mechanics, so it fits here). The paragraph is now just: same idea as SBOM, but for cryptography, with the TLS/cipher-suite example.

docs/SUPPLY_CHAIN_SECURITY.md
brawer: Fair on both counts. Went with "It's also about where things are heading, not just where they stand today" — drops the competitive "bet" framing and the corporate-sounding "industry", and is a touch shorter too.

docs/SUPPLY_CHAIN_SECURITY.md
```diff
@@ -1,192 +1,170 @@
...
-
-It’s also a reasonable bet on where the industry is heading, not just
-where it stands today: supply-chain attacks aren’t hypothetical anymore
-(the [xz-utils backdoor](https://en.wikipedia.org/wiki/XZ_Utils_backdoor)
-being the starkest recent example), and the regulation and tooling
-responding to them — the practices this document describes among them —
-are becoming baseline expectations rather than something only
+Fair question. `osm-diffs` is open source, processes only public data,
+and handles no secrets — our actual threat model is modest compared to
+a proprietary service handling user data or credentials. So no, this
```
brawer: Remove "proprietary"

brawer: Done — dropped it. The contrast being made is about what a service handles (user data, credentials), not who owns it, so "proprietary" was not adding anything, just an unintended swipe.

brawer: Exactly, thanks.

PR #599, opened 2026-08-11
ci: cancel superseded test runs via a concurrency group

brawer:
## What

Add a `concurrency` group to `Tests` so a newer run for the same PR (or
branch/merge-queue entry) cancels a still-running older one, instead of
queuing behind or running alongside it.

## Why

Noticed on PR #598: pushing a merge commit ("Update branch") to a PR
started a second run while the first was still in progress. It just sat
`queued` behind the older run rather than being cancelled, wasting a
runner slot and delaying feedback on the actually-current commit.

Also added a comment clarifying that this workflow never triggers on tags
(the `push` trigger is restricted to `main`) — tag pushes are handled by
`release.yml` (see `docs/RELEASING.md` and
`docs/SUPPLY_CHAIN_SECURITY.md`), so `github.ref`, used in the concurrency
group key, is always a branch ref here, never a tag ref.

Independent of #597/#598 — unrelated to the build-caching work, just CI
hygiene noticed along the way.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,5 +1,11 @@
 name: Tests
 
+# This workflow deliberately never triggers on tags: the push trigger below
+# is restricted to the "main" branch. Pushing a git tag is handled by
+# release.yml instead; see docs/RELEASING.md and docs/SUPPLY_CHAIN_SECURITY.md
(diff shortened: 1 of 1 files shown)


brawer (commented):
LGTM

PR #603, opened 2026-08-11
Split release.yml into caller + reusable workflow: reach SLSA Build Level 3

brawer:
Follow-up to #601. Per GitHub's own documented compliance bar ([source](https://github.com/github/docs/blob/main/content/actions/how-tos/secure-your-work/use-artifact-attestations/increase-security-rating.md)), `actions/attest` used directly — even isolated into its own job, as we already had — is Build Level 2. Reaching Level 3 specifically requires moving build *and* attestation into a reusable workflow, so the process that generates provenance has a distinct, independently verifiable identity, rather than being just another job in the same maintainer-editable file the build steps live in.

## What changed

- **`release-build.yml`** (new): a reusable workflow (`on: workflow_call`) holding `verify-version`/`build`/`manifest`/`attest` — moved verbatim from `release.yml` (confirmed via `diff` against the old file: only the old attest-job comment was replaced by a new file-header explaining why this file has to stay separate).
- **`release.yml`**: now a thin caller — the tag-push trigger, the `release`/`:latest` concurrency group (unchanged; it governs the whole run regardless of what's inside, so the serialization guarantee against two releases racing on `:latest` still holds), and one job that calls `release-build.yml` with the union of permissions its jobs need.
- **`scripts/verify-release.sh`**: now checks this for real. It reads a real provenance attestation's `predicate.buildDefinition.externalParameters.workflow.path` and fails loudly if it isn't `.github/workflows/release-build.yml` — so every future release run re-verifies the SLSA L3 property empirically, not just by trusting that the file split does what it's supposed to.
- **`docs/RELEASING.md`**: "Where things live" now lists both files.

`docs/SUPPLY_CHAIN_SECURITY.md` is deliberately *not* updated to claim Level 3 in this PR — that's the planned last step, once this has actually been exercised by a real release and `verify-release.sh`'s new check has passed for real.

## Things I verified before relying on them

- **`github` context and `secrets.GITHUB_TOKEN` propagate automatically** to a called reusable workflow without needing `inputs`/`secrets: inherit` — confirmed against GitHub's own reference docs (`reusing-workflow-configurations.md`): *"When a reusable workflow is triggered by a caller workflow, the `github` context is always associated with the caller workflow. The called workflow is automatically granted access to `github.token` and `secrets.GITHUB_TOKEN`."*
- **Permissions only flow down, never up** — confirmed the caller job's `permissions:` block is a superset of every job's needs inside `release-build.yml` (`contents: read`, `id-token: write`, `packages: write`, `attestations: write`, `artifact-metadata: write`).
- **The recorded builder identity actually is the called workflow, not the caller** — pulled a real provenance attestation from the existing v0.6.9 release and inspected its actual JSON shape (`predicate.buildDefinition.externalParameters.workflow.path`, `predicate.runDetails.builder.id`) rather than guessing field names for the new `verify-release.sh` check.
- **`verify-release.sh`'s existing `gh run list --workflow=release.yml`/`gh run download` calls need no changes** — `release.yml` is still the top-level triggered workflow name, and jobs from a called reusable workflow share the same run ID (same run, nested job group in the Actions UI), so run-lookup and artifact download are unaffected by the split.

## Testing

- `actionlint` (workflow + shellcheck) on both new/changed workflow files: identical finding count to the pre-split file (24 pre-existing shellcheck style warnings, none new).
- `diff` of the job bodies between the old `release.yml` and new `release-build.yml`: byte-identical apart from the intentionally replaced header comment.
- Ruby's YAML parser (`YAML.load_file`) on both workflow files: valid.
- `sh -n` on the updated `verify-release.sh`: valid; no `shellcheck` findings.
- **Not yet tested**: an actual tag push through this new structure — that can only be verified by cutting a real release, which I'd suggest doing as the next one. If `verify-release.sh`'s new builder-path check fails, that means the isolation assumption above didn't hold in practice and needs debugging before we can honestly claim L3.

### DIFF
--- a/.github/workflows/release-build.yml
+++ b/.github/workflows/release-build.yml
@@ -1,0 +1,263 @@
+name: Build and attest a release
+
+# Reusable workflow, called only by release.yml. Kept as a separate workflow
+# file (rather than a job in release.yml) specifically so this process has
+# its own, distinct identity for provenance purposes: the attestations this
+# workflow signs record *this* file's path and pinned ref as the builder,
+# not release.yml's. That's what SLSA Build Level 3 requires -- see
+# "Build provenance and attestations" in docs/SUPPLY_CHAIN_SECURITY.md for
+# why, and don't fold these jobs back into release.yml.
+# https://slsa.dev/spec/v1.2/build-track-basics#build-l3
+on:
+  workflow_call:
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+
+permissions: {}
+
+jobs:
+  verify-version:
+    name: Verify release tag matches Cargo.toml version
+    runs-on: ubuntu-24.04
+    timeout-minutes: 5
+    permissions:
+      contents: read
+
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
+
+      - name: Compare tag to Cargo.toml version
+        run: |
+          TAG="${{ github.ref_name }}"
+          CARGO_VERSION=$(awk '
+            /^\[package\]/ { in_package=1; next }
+            /^\[/           { in_package=0 }
+            in_package && /^version[[:space:]]*=/ {
+              match($0, /"[^"]*"/)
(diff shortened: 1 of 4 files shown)


brawer (commented):
LGTM

PR #608, opened 2026-08-11
verify-release.sh: check runDetails.builder.id, not externalParameters

brawer:
Cut v0.6.10 as the first real test of the `release-build.yml` split (#603) and the new builder check in `verify-release.sh` failed:

```
ERROR: provenance for amd64 was signed by workflow
'.github/workflows/release.yml', expected
'.github/workflows/release-build.yml'.
```

## What actually happened

Pulled the real attestation JSON for v0.6.10 to find out why. Good news: **the split itself works exactly as intended.** `predicate.runDetails.builder.id` correctly says `https://github.com/alltheplaces/osm-diffs/.github/workflows/release-build.yml@refs/tags/v0.6.10` — confirmed identical for the amd64, arm64, and manifest attestations.

The check was just reading the wrong field. `predicate.buildDefinition.externalParameters.workflow.path` records the *caller* workflow that responded to the tag-push event (`release.yml`) — that's correct and expected, it just isn't "who signed this." `predicate.runDetails.builder.id` is the actual SLSA builder identity, and that one does say `release-build.yml`.

## Also checked (per a request to look at the run's warnings)

Two warnings appear in the v0.6.10 run (`WARNING: image platform ... does not match the expected platform` in the manifest job, and `Failed to create storage record` / `artifact-metadata:write` in the attest job). Compared against the pre-split v0.6.9 run: both are byte-identical there too, including the storage-record warning's eventual self-resolving retry-then-succeed pattern. Pre-existing GitHub/podman quirks, unrelated to this migration — no action needed.

## Testing

Ran the fixed script against the real v0.6.10 release (not a synthetic/mocked test):

```
$ ./scripts/verify-release.sh v0.6.10
...
v0.6.10 verified: release.yml succeeded, and both SBOM and build-provenance attestations exist for both architectures.
```

Passes cleanly. This is real, empirical confirmation that `release.yml`/`release-build.yml` (#603) actually achieves SLSA Build Level 3 — not just that the file structure looks right.

## Next

With this confirmed, I'll follow up with the doc update to `docs/SUPPLY_CHAIN_SECURITY.md` claiming Level 3 for real (the step that was deliberately deferred in #603 until this was verified).

### DIFF
--- a/scripts/verify-release.sh
+++ b/scripts/verify-release.sh
@@ -83,12 +83,18 @@ ARM64_DIGEST="$(cat "${WORKDIR}/arm64/digest.txt")"
 # Note: `gh attestation verify`'s default --predicate-type filter only shows
 # build provenance, hiding the SBOM attestation; query the raw API instead.
 #
-# Also confirms the provenance was signed by release-build.yml, not
-# release.yml: that's the actual, checkable evidence that build+attest ran
-# inside the isolated reusable workflow SLSA Build Level 3 requires, rather
-# than just trusting the workflow file's own claim about itself. See
-# "Build provenance and attestations" in docs/SUPPLY_CHAIN_SECURITY.md.
-EXPECTED_BUILDER_PATH=".github/workflows/release-build.yml"
+# Also confirms the provenance's *builder* (predicate.runDetails.builder.id)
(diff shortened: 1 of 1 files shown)


brawer (commented):
LGTM

PR #609, opened 2026-08-11
docs/SUPPLY_CHAIN_SECURITY.md: document that we now reach SLSA Build L3

brawer:
Last step of the plan from #601: correct the claim to L2 (#601) → do the actual reusable-workflow split (#603) → fix `verify-release.sh`'s builder check and confirm it against a real release (#608, v0.6.10) → **document L3 for real (this PR)**.

Deliberately not folded into #603 — a workflow split that looks structurally right isn't the same as one that's been proven to work. v0.6.10 is that proof: `verify-release.sh` read the real attestation's `predicate.runDetails.builder.id` for v0.6.10 and confirmed it's `release-build.yml`, not `release.yml`.

## Testing

- `grep -rn "Build Level 2\|Build Level 3" docs .github` — no stale L2 claims remain anywhere; the `release.yml`/`release-build.yml` header comments (written in #603) were already accurate and needed no change.
- No straight quotes/dashes crept back in (`grep -n '"'` / em-dash spot check both clean).
- 1230 words (was 1217 after the last tightening pass, #596) — the added paragraph explaining *why* we can trust the builder identity is worth the length here, this is the section's central claim.

### DIFF
--- a/docs/SUPPLY_CHAIN_SECURITY.md
+++ b/docs/SUPPLY_CHAIN_SECURITY.md
@@ -108,19 +108,20 @@ workflow W, from commit C, on GitHub-hosted infrastructure, triggered
 by event E.”
 
 We reach
-[SLSA Build Level 2](https://slsa.dev/spec/v1.2/build-track-basics#build-l2):
-provenance that’s signed by a hosted build platform (GitHub Actions),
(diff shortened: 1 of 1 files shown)


brawer (commented):
LGTM

PR #610, opened 2026-08-11
ci: note why sccache is kept alongside the target/ cache

brawer:
Comment-only change, no behavior change. Records the measured reason sccache stays alongside `Swatinem/rust-cache` (~180s/PR saved, per the A/B experiment in #607) directly in the workflow, so it's visible without digging through #597's comments.

Closes #597 (last item).


brawer (commented):
LGTM

PR #611, opened 2026-08-11
docs/SUPPLY_CHAIN_SECURITY.md: shorten the SLSA L3 paragraph

brawer:
The L3 paragraph had grown into an implementation walkthrough — `predicate.runDetails.builder.id`, "an attacker who could edit `release.yml`", exactly what `verify-release.sh` checks. That's real information, but this is meant to be a conceptual, high-level doc: a reader new to release engineering doesn't need the field name to get the point, and the mechanism detail was duplicating what `release-build.yml`'s own header comment and `verify-release.sh` already explain at the depth an implementer actually wants.

Cut it down to the concept — build-triggering and signing have separate, independently checkable identities — and linked out for the receipts instead: `release-build.yml` and `verify-release.sh` for the mechanism, #608 for this having been checked against a real release rather than assumed. Matches how the multi-arch nuance a few paragraphs down already links out to #589 rather than explaining inline.

**1230 → 1198 words.**

Scanned the rest of the doc for further cuts too (this was requested as "another round" of shortening), but didn't find flabby prose worth touching elsewhere — the other sections already went through this same tightening pass in #596, and forcing more cuts there risked losing clarity for no real word-count gain.

### DIFF
--- a/docs/SUPPLY_CHAIN_SECURITY.md
+++ b/docs/SUPPLY_CHAIN_SECURITY.md
@@ -109,19 +109,17 @@ by event E.”
 
 We reach
 [SLSA Build Level 3](https://slsa.dev/spec/v1.2/build-track-basics#build-l3):
-provenance whose builder identity is distinct from, and independently
-verifiable from, whatever triggered the build. `release.yml` only
-triggers on a tag push and calls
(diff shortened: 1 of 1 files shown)


brawer (commented):
Nice, thanks for the rewrite. LGTM

PR #612, opened 2026-08-11
release-build.yml: fix pre-existing shellcheck warnings (via actionlint)

brawer:
Fixes all 24 `shellcheck` findings that `actionlint` reports in the `build`/`manifest`/`attest` jobs of `.github/workflows/release-build.yml`:

- **SC2006** (×7) — legacy backticks replaced with `$(...)`
- **SC2046** (×7) — command substitutions quoted to prevent word splitting
- **SC2086** (×7) — variables sourced from file/command output quoted
- **SC2034** (×2) — removed the dead `VCS_REF`/`VCS_URL` assignments in the `manifest` job (they're only needed as `--build-arg`s in the separate `build` job)
- **SC2129** (×1) — consolidated three individual `>> $GITHUB_OUTPUT` redirects into one `{ ...; } >> "$GITHUB_OUTPUT"` block in the `attest` job

Not urgent, none were exploitable (no untrusted input reaches these strings) — pure cleanup.

Verified locally:
```
$ actionlint .github/workflows/release-build.yml
$ echo $?
0
```

Fixes #604

### DIFF
--- a/.github/workflows/release-build.yml
+++ b/.github/workflows/release-build.yml
@@ -89,7 +89,7 @@ jobs:
         id: commit-info
         run: |
           TIMESTAMP=$(git log -1 --format=%cI ${{ github.sha }})
-          echo "timestamp=$TIMESTAMP" >> $GITHUB_OUTPUT
+          echo "timestamp=$TIMESTAMP" >> "$GITHUB_OUTPUT"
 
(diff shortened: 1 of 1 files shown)


brawer (commented):
LGTM

PR #614, opened 2026-08-11
Remove fuzz/ target for GeoJSON import

### LINKED ISSUE #91: Set up fuzzing for GeoJSON parser

brawer:
Closes #91.

The `cargo-fuzz` target for `import_atp`'s GeoJSON parsing has never
been wired into CI, and had bit-rotted without anyone noticing: the
`#[cfg(fuzzing)]` re-export in `src/lib.rs` pointed at
`import_atp::fuzz::...`, a module path that stopped existing once the
`atp` module was renamed — so the fuzz target hasn't actually compiled
for a while.

Rationale for dropping it rather than fixing it, from the discussion
on #91: the GeoJSON is parsed with the well-tested `geojson`/
`serde_json` crates and the parsing code already treats malformed
input as data, not danger (`Result`-based errors, no unwraps on
attacker-shaped input), so the main thing fuzzing would realistically
catch — a stray panic — is already unlikely and already handled. The
crash-regression unit test that exercised the fuzz entry point is kept,
folded into the regular `atp::tests` module.


brawer (commented):
LGTM

PR #617, opened 2026-08-11
Add src/memstats.rs: process/cgroup memory snapshots

brawer:
Part of #616.

Standalone module exposing `memstats::snapshot()`: current + peak RSS
with its anon/file/shmem breakdown from `/proc/self/status`, and cgroup
v2 `memory.current`/`memory.max`/`memory.peak` from `/sys/fs/cgroup`, on
Linux; peak RSS via `getrusage()` on macOS. See the module doc comment
for the full reasoning (why these fields, why `RssFile` vs. `RssShmem`
matters for `src/tables`'s mmap'd files, why macOS reads `ru_maxrss`
rather than `phys_footprint`).

Not wired into the pipeline yet -- that's the next PR in this stack, so
this one is marked `#[allow(unused)]` for now, matching the existing
pattern in `src/tables/mod.rs` for landing infrastructure ahead of its
use.

`libc` is added as a **macOS-only** target dependency (`getrusage` is
the only place that needs FFI; the Linux path is pure file parsing), so
it doesn't touch the Linux/musl production build. It was already a
transitive dependency, so nothing new enters the dependency tree/SBOM.

--locked` (all 205+ tests pass), `cargo clippy --all-targets -- -D
warnings` -- all clean.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2712,6 +2712,7 @@ dependencies = [
  "geo-traits",
  "geojson",
  "indicatif",
+ "libc",
  "librqbit",
  "log",
  "lru",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,6 +45,12 @@ tokio = { version = "1.52.3", default-features = false }
 webpki-roots = "1.0.6"
 wkb = "0.9.2"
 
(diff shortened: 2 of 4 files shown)


brawer (commented):
LGTM

PR #618, opened 2026-08-11
pipeline: log a memory snapshot at the start/end of each step

brawer:
Part of #616. Stacked on #617 -- this PR only makes sense once that one
lands; the diff here is against `add-memstats-module`, not `main`.

Wraps each of the 7 top-level steps in `run_pipeline()` with a small
`run_step()` helper that logs a `crate::memstats::snapshot()` before and
after -- on both success and failure, so a step that errors out (e.g. an
actual OOM) still gets its "end" snapshot on the way out.

Verified end-to-end against the real binary using the existing
integration test's fixtures, e.g.:

```
import_atp: start rss_peak_bytes=14680064
import_atp: end rss_peak_bytes=32817152
build_coverage: start rss_peak_bytes=32817152
build_coverage: end rss_peak_bytes=35766272
...
```

(On this macOS dev machine only `rss_peak_bytes` is populated; in the
Linux/OCI production container the `rss_*`/`cgroup_*` breakdown from
`/proc` and `/sys/fs/cgroup` would show up too.)

--locked` (207 tests, all pass, including the pre-existing end-to-end
`test_pipeline` integration test), `cargo clippy --all-targets -- -D
warnings` -- all clean.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -6,8 +6,6 @@ mod coverage;
 mod geometry;
 mod logging;
 mod matchers;
-// Not yet wired into the pipeline; see #616.
-#[allow(unused)]
 mod memstats;
 mod pipeline;
 mod places;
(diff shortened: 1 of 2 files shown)


brawer (commented):
LGTM

PR #619, opened 2026-08-11
pipeline: warn when nearing the cgroup memory limit

brawer:
Part of #616. Stacked on #618 (which is stacked on #617) -- diff here is
against `log-pipeline-step-memory`, not `main`.

Adds `MemStats::cgroup_usage_fraction()` (`current / max`, `None` if
either is unavailable or the limit is unconfigured), and has the
per-step snapshot logging emit a `log::warn!` once usage crosses 85% of
the cgroup limit -- an early signal ahead of the kernel's actual
OOM-kill at 100%.

Also answers a question that came up while reviewing this: what do the
`cgroup_*` fields report on Linux outside a container (a VM or bare
metal)? Documented in the module doc comment -- short version: they
usually still populate under systemd (every session/service gets its
own cgroup), `cgroup_max_bytes` correctly comes back `None` absent an
explicit `MemoryMax=` rather than reporting something misleading, and
`cgroup_current_bytes`'s scope is whatever cgroup the process happens to
be in (tight if run as its own systemd service, broader if run by hand
in an interactive shell). No code changes were needed for this --
the existing "unavailable sources are `None`, never an error" design
already covers it.

warnings` -- all clean.

### DIFF
--- a/src/memstats.rs
+++ b/src/memstats.rs
@@ -21,6 +21,20 @@
 //! kill a pod" -- but they don't exist on macOS, hence the
 //! [`MemStats::rss_peak_bytes`] fallback, which is available everywhere.
 //!
+//! On Linux outside a container -- a VM or bare metal -- the `cgroup_*`
+//! fields usually still populate: systemd puts every login session and
+//! service unit into its own cgroup, so `/sys/fs/cgroup/memory.current`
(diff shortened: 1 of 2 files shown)


brawer (commented):
LGTM

PR #622, opened 2026-08-11
logging: support structured key-value fields in the JSON logger

brawer:
Part of #621.

Enables `log`'s `kv` feature and extends `format_json` to render a
record's `key_values()` into a `"fields"` JSON object, via a small
`VisitSource`/`VisitValue` adapter. Deliberately not the `kv_serde`
feature (and the `serde` dependency on `log` that comes with it): the
value types actually needed (`u64`, `f64`, `&str`, `bool`, and `Option`
of those) are already covered by `VisitValue`'s primitive methods
directly.

Purely additive: records that don't attach any key-value pairs -- every
existing `log::info!`/`log::warn!` call site elsewhere in the crate,
unchanged -- get no `"fields"` key at all. An absent `Option` field
serializes as an explicit JSON `null` at a fixed key rather than being
omitted, which is more useful for a structured consumer (`jq`, a log
aggregator) than a variably-shaped object; both this and a
str/u64/f64/bool round-trip are covered in the extended test.

This PR is the mechanism only -- it deliberately doesn't retrofit
existing message-text call sites elsewhere in the codebase. The next PR
in this stack (#621) uses it for per-step elapsed time and switches the
pipeline's memory-snapshot logging over to it.

warnings` -- all clean.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ geo-traits = { version = "0.3.0", default-features = false }
 geojson = "1.0.0"
 indicatif = { version = "0.18.4", default-features = false }
 librqbit = { version = "8.1.1", default-features = false, features = ["disable-upload", "rust-tls"] }
-log = { version = "0.4.32", default-features = false }
+log = { version = "0.4.32", default-features = false, features = ["kv"] }
 lru = { version = "0.18.0", default-features = false }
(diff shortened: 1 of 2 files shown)


brawer (commented):
LGTM

PR #623, opened 2026-08-11
pipeline: log step elapsed time via structured fields

brawer:
Part of #621. Stacked on #622 -- diff here is against `logging-kv-fields`,
not `main`.

`run_step` now measures each step's wall-clock run time with `Instant`,
and `log_snapshot` moves from interpolating `MemStats`'s `Display` into
the message text to attaching `step`, `phase`, `elapsed_seconds` (`None`
on "start", the run time on "end"), and every `MemStats` field as real
`log::kv` fields -- queryable directly (`jq '.fields.elapsed_seconds'`)
rather than parsed out of a string. The cgroup-limit warning is
restructured the same way.

Verified end-to-end against the real binary, e.g.:

```json
{"fields":{"step":"import_osm","phase":"end","elapsed_seconds":0.152129292,
 "rss_peak_bytes":43188224,"cgroup_current_bytes":null, "cgroup_max_bytes":null,
 "cgroup_peak_bytes":null,"rss_bytes":null,"rss_anon_bytes":null,
 "rss_file_bytes":null,"rss_shmem_bytes":null},
 "level":"INFO","message":"import_osm: end","target":"osm_diffs::pipeline", ...}
```

(macOS fields are `null` as expected; a Linux/production run would
additionally populate the `rss_*`/`cgroup_*` breakdown.)

warnings` -- all clean.

### DIFF
--- a/src/pipeline/mod.rs
+++ b/src/pipeline/mod.rs
@@ -1,5 +1,8 @@
 use anyhow::{Context, Ok, Result};
-use std::{path::Path, time::SystemTime};
+use std::{
+    path::Path,
+    time::{Instant, SystemTime},
+};
 
 mod conflate;
 mod edits;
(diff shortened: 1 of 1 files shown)


brawer (commented):
LGTM

PR #624, opened 2026-08-11
Extract shared sorted-u64-key lookup into src/tables/sorted_u64_index.rs

brawer:
## What

`BlobTable::lookup`, `CoordTable::get`, and `U64Set::contains` each duplicated the same binary search over a sorted, mmap'd, little-endian `&[u64]` slice. This pulls that logic into one shared `src/tables/sorted_u64_index.rs::search`, used by all three. No behavior or performance change.

## Why

Prompted by looking into whether `CoordTable` (286.7M entries on a full planet import) would benefit from a faster lookup than binary search, given OSM node IDs are mostly contiguous. I prototyped a guarded interpolation search, but a benchmark against a real mmap'd file (not a heap `Vec`, which turned out to hide the effect) showed it was consistently ~33–35% slower than binary search once the backing pages were cache-resident — true for nearly every run. Full writeup and numbers in #620, which I'm closing as not pursued for now.

What's left from that investigation is this dedup, which stands on its own.

## Testing

- `cargo test` (unit + integration + doctests): all pass
- `cargo clippy --lib --tests --all-targets`: clean
- `cargo fmt --check`: clean

### DIFF
--- a/src/tables/blob_table.rs
+++ b/src/tables/blob_table.rs
@@ -125,14 +125,10 @@ impl<'a> BlobTable<'a> {
     }
 
     pub fn lookup(&self, key: u64) -> Option<&'a [u8]> {
-        let idx = self.keys.partition_point(|&k| u64::from_le(k) < key);
-        if idx < self.keys.len() && u64::from_le(self.keys[idx]) == key {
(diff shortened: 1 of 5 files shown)


brawer (commented):
LGTM

PR #628, opened 2026-08-12
Fall back to build_line for closed ways too degenerate to be a ring

brawer:
## Problem

`is_area()`'s tag heuristic can misfire on a closed way that's really a degenerate line — e.g. `way/1464457914`, a `man_made=tunnel`/`tunnel=culvert` whose start node was reused as its end node, "closing" a path with only one interior node (`A → B → A`) and thus no possible area. `build_ring` correctly rejects that (there's no polygon to build), and the feature was dropped outright with `could not build geometry for way/…`.

Full analysis in #627.

## Fix

A closed ring needs at least 3 distinct vertices — 4 raw node refs, counting the repeated closing one — to possibly enclose any area. Below that, `build_ring` is guaranteed to return `None` no matter what `is_area()` says. So route those straight to `build_line` instead:

```rust
fn ring_is_geometrically_possible(way_members_count: usize) -> bool {
    way_members_count >= 4
}
```

This adds no cost on the common path — no clone, no wasted `build_ring` call — since the check is purely a count comparison done before deciding which builder to call.

## Testing

- Added `ring_is_geometrically_possible_requires_at_least_four_refs`, pinning the boundary at 4 (now load-bearing for correctness).
- `cargo test --lib`: 217 passed.
- `cargo fmt --check` / `cargo clippy --all-targets`: clean.

Fixes #627

### DIFF
--- a/src/pipeline/osm/assemble.rs
+++ b/src/pipeline/osm/assemble.rs
@@ -274,7 +274,17 @@ fn assemble_ways<'a>(
                         // a Point, LineString, Polygon, MultiLineString, or MultiPolygon.
                         let is_closed = way_members_count >= 2
                             && way_members[0] == way_members[way_members_count - 1];
-                        let geometry = if is_area(is_closed, way.tags()) {
+                        // `is_area()`'s tag heuristic can misfire on a closed way that's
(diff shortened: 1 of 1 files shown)


brawer (commented):
LGTM

PR #631, opened 2026-08-12
Add centroid_s2_cell_id spatial sort key to FeatureToIndex

brawer:
## What

Adds `fixed64 centroid_s2_cell_id` to `FeatureToIndex`: the S2 cell ID of the feature geometry's centroid. Sorting numerically by this key clusters spatially nearby features together, since S2 cell IDs are ordered along a Hilbert curve — this is the sort key we'll use to spatially sort the index.

## Details

- `FeatureToIndex` is purely internal/ephemeral (only ever lives under the pipeline's scratch working directory, never sent over the wire or stored long-term), so all fields were renumbered starting from 1 instead of appending the new field at the end.
- `assemble_geometry()` now takes `&mut FeatureToIndex` instead of `&mut Vec<u64>`, and derives `centroid_s2_cell_id` from the same centroid computation already needed for some geometry kinds — so all four call sites (nodes, ways, leaf relations, super relations) get both the new sort key and the existing coverage cell IDs in a single pass, no redundant centroid computation. This also resolves the `TODO: also store the s2 cell of the centroid` left in the super-relations path.
- Renamed the existing `repeated uint64 s2_cell_id` field to `coverage_s2_cell_id`: next to `centroid_s2_cell_id`, the old name no longer said what it actually holds — an approximate covering set of cells for spatial matching, not a single sort key.

## Testing

- `cargo test --lib`: 217 passed.
- `cargo fmt --check` / `cargo clippy --all-targets`: clean.

### DIFF
--- a/src/tables/feature.proto
+++ b/src/tables/feature.proto
@@ -43,7 +43,7 @@ message FeatureToIndex {
 
     // A set of S2 cell IDs that cover the geometry of the indexed feature,
     // for approximate spatial matching. Unlike `centroid_s2_cell_id`, this
-    // isn't a single point and isn't meant for sorting.
+    // isn't necessarily a single point and isn't meant for sorting.
     repeated uint64 coverage_s2_cell_id = 4;
 
     // Here would be a good place to store supplemental information
--- a/src/tables/feature.proto
+++ b/src/tables/feature.proto
@@ -24,17 +24,27 @@ message RelationMember {
   uint64 id = 2;
 }
 
-// An OpenStreetMap feature to be indexed.
+// An OpenStreetMap feature to be indexed. This is an internal, ephemeral
+// data structure: it only ever lives in files under the pipeline's scratch
+// working directory, and is never sent over the wire or stored long-term.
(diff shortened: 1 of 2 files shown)


src/tables/feature.proto
```diff
@@ -24,17 +24,27 @@ message RelationMember {
...
     // A bitmask to quickly skip over features that cannot possibly match
     // the query feature. See matchers::MatchMask.
-    uint32 match_mask = 2;
+    uint32 match_mask = 3;
 
-    // A set of S2 cell IDs that cover the geometry of the indexed feature.
-    repeated uint64 s2_cell_id = 3;
+    // A set of S2 cell IDs that cover the geometry of the indexed feature,
+    // for approximate spatial matching. Unlike `centroid_s2_cell_id`, this
+    // isn't a single point and isn't meant for sorting.
```
brawer: Change "isn't a single point" to "isn't necessarily a single point". For nodes (or ways and relations whose geometry is a degenerate line or polygon), this absolutely will be a single point.

brawer (commented):
LGTM

PR #632, opened 2026-08-12
Document Feature and RelationMember fields in feature.proto

brawer:
## What

Comment-only change: documents every field of `Feature` and `RelationMember` in `feature.proto`. Prompted by reviewing doc clarity after #631 landed `centroid_s2_cell_id`/`coverage_s2_cell_id` on `FeatureToIndex` — that message ended up well-documented, but `Feature`/`RelationMember` had real gaps.

## Gaps fixed

- **`tags`**: gave no clue it's flattened `(key_id, value_id)` `StringPool`-index pairs. That fact only existed in a private helper's rustdoc in `assemble.rs`, invisible from the proto itself.
- **`way_members` vs `RelationMember.id`**: both `uint64`, one message apart, look parallel — but `way_members` holds raw OSM node IDs while `RelationMember.id` holds `id`-encoded feature IDs (`osm_id*10 + {1,2,3}`). Nothing warned a reader they're differently encoded.
- **`RelationMember.role`**: no comment at all — not obviously a `StringPool` index rather than an enum or the literal role string.
- **`version`/`changeset`/`timestamp`**: undocumented, no units, no note that they default to `0` when OSM doesn't supply them.
- **`geometry_wkb`**: didn't say it's little-endian WKB with (longitude, latitude) coordinate order.
- No message-level note on which fields apply to which OSM element kind (`way_members` ways-only, `relation_members` relations-only).

Also gave `Feature` the same "internal/ephemeral, never sent over the wire or stored long-term, field numbers are free to be renumbered" note `FeatureToIndex` already had. Since `Feature` is declared first in the file, the full explanation lives there and `FeatureToIndex` just refers back to it, so a top-to-bottom reader gets the full story once, in the right order.

Fixed two pre-existing typos in `FeatureToIndex`'s trailing comment (`should't` → `shouldn't`, `finalm` → `final`) while in the area.

All field semantics were checked against the actual population code (`assemble_tags`, `assemble_relation_members`, `assemble_feature` in `assemble.rs`), not guessed.

## Testing

`cargo build` (regenerates the prost bindings from the proto) and `cargo fmt --check`: clean. No `.rs` changes, so no behavior change.

### DIFF
--- a/src/tables/feature.proto
+++ b/src/tables/feature.proto
@@ -5,29 +5,61 @@ syntax = "proto3";
 
 package tables.features;
 
-// An OpenStreetMap feature stored by a FeatureTable.
+// An OpenStreetMap feature (node, way, or relation) stored by a
+// FeatureTable, keyed by `id`. `way_members` is only populated for ways,
+// and `relation_members` only for relations.
+//
+// This is an internal, ephemeral data structure: it only ever lives in
+// files under the pipeline's scratch working directory, and is never sent
+// over the wire or stored long-term. Field numbers are free to be
+// renumbered as the message evolves.
 message Feature {
     // id = (osm_id * 10) + 1 for nodes; + 2 for ways; + 3 for relations.
     uint64 id = 1;
+
+    // OSM edit version of this element, as assigned by OpenStreetMap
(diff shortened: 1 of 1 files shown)


brawer (commented):
LGTM

PR #634, opened 2026-08-12
Expose OSMHeader provenance metadata from BlobReader

brawer:
## Summary
- `BlobReader` now parses the `OSMHeader` block's `writingprogram`, `source`, and `osmosis_replication_timestamp` fields by hand (neither `osm_pbf_iter` nor `protobuf_iter` support that message) and exposes them via a new `BlobReader::header() -> &PbfHeader`.
- `replication_timestamp` is required: `BlobReader::open()` now errors if the header block is missing or has no timestamp. In practice this never happens for the planet dumps we process. `replication_sequence_number` is intentionally not exposed — `planet-dump-ng`, the tool behind the actual `planet.openstreetmap.org` dumps, never sets it in the header.
- `import_osm` logs the parsed metadata at INFO level (structured, via `log`'s key-value API) right after opening the planet file, so provenance is visible in `pipeline.log` ahead of a future change to embed it into our output files.

## Why
We want to eventually stamp our output files with the provenance of the input planet data (replication timestamp etc.). This is the first step: get `BlobReader` to expose it and log it. Embedding it into output files is a follow-up.

## Testing
- `cargo test --lib pipeline::osm::tests` and `cargo test --test integration_test` pass.
- `cargo fmt --check` and `cargo clippy --lib -- -D warnings` are clean.
- Manually verified the logged JSON line against the `zugerland.osm.pbf` test fixture (not the real planet dump):
  ```json
  {"fields":{"replication_timestamp":"2026-01-27T08:11:02Z","source":null,"writing_program":"osmx"},"level":"INFO","message":"opened OpenStreetMap planet file", ...}
  ```
- Cross-checked the field parsing (tags, wire types, real values) against both the checked-in test fixture and the real planet dump in `workdir/osm-planet.pbf` by hand-decoding the header block with a throwaway script, to confirm `osmosis_replication_sequence_number` really is absent from real planet dumps and not just from the small test fixture.

### DIFF
--- a/src/pipeline/osm/mod.rs
+++ b/src/pipeline/osm/mod.rs
@@ -12,6 +12,8 @@ use std::ops::Range;
 use std::path::{Path, PathBuf};
 use std::sync::mpsc::{SyncSender, sync_channel};
 use std::thread;
+use time::UtcDateTime;
+use time::format_description::well_known::Rfc3339;
 
 use crate::coverage::Coverage;
 use crate::make_progress_bar;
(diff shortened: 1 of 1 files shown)


src/pipeline/osm/mod.rs
```diff
@@ -429,6 +466,53 @@ impl<'a, R: Read + Seek + Send> BlobReader<'a, R> {
...
     }
+
+    /// Parses the fields we care about out of a decoded `OSMHeader` blob,
+    /// i.e. the `HeaderBlock` message from OSM's `osmformat.proto`.
+    /// `osm_pbf_iter`/`protobuf_iter` don't support this message, so we
+    /// pick out the fields we need by hand:
+    ///
+    /// - `writingprogram` (tag 16, string)
+    /// - `source` (tag 17, string)
+    /// - `osmosis_replication_timestamp` (tag 32, int64)
```
brawer, line 477: Also explain why we don't parse osmosis_replication_sequence_number, summarizing what you discovered during our discussion. Please include the timestamp (as human-readable RFC3339 string) of the OSM planet file that you've analyzed. (No need to mention the testdata file).

brawer (commented):
LGTM

PR #640, opened 2026-08-12
Capture and log AllThePlaces run provenance metadata

brawer:
## Summary
- New `AtpMetadata` in `src/atp/fetch.rs`, mirroring `PbfHeader` from #634 but for the AllThePlaces run that feeds `alltheplaces.parquet`: `run_id`, `output_url`, `history_url`, `start_time`, `end_time`, `spiders`, `total_lines`, `size_bytes`, all parsed from AllThePlaces' `history.json`.
- All fields are required. `history.json` only lists finished runs, so `fetch_latest_run` walks entries from newest to oldest, skipping (and `WARN`-logging) any entry that's missing a field or whose stats look broken -- below `MIN_SPIDERS`/`MIN_TOTAL_LINES`/`MIN_SIZE_BYTES`, deliberately loose floors just meant to catch a catastrophically broken run, not routine fluctuation. If the newest usable entry is older than `MAX_STALENESS` (6 weeks), we refuse to run rather than silently work from stale input.
- `import_atp` logs the metadata at INFO right after fetching -- not CycloneDX yet, just the underlying attributes -- so it's visible in `pipeline.log` ahead of the actual embedding work tracked in #638.
- `fetch_atp`'s cached-zip path now reads the metadata back from `alltheplaces.meta.json` instead of silently skipping it, so provenance survives a re-run that reuses an already-downloaded zip.
- Bonus fix found along the way: `run_step` (`src/pipeline/mod.rs`) already logged a memstats/timing snapshot whether a step succeeded or failed, but never the error itself -- so a fatal error only ever reached `pipeline.log` as an unexplained "end" snapshot, with the actual message only surfacing via `main()`'s default unwind on process exit. Now `run_step` logs it at ERROR with the full `anyhow` context chain, in the one place every top-level step already passes through.

## Why
Closes #639, a sub-issue of #638 (embedding pipeline-input provenance into our output as a CycloneDX-style data BOM). This is the AllThePlaces analogue of #634: parse the metadata and log it, ahead of the actual embedding work.

## Testing
- `cargo test --lib`: 220 passed.
- `cargo fmt --check` and `cargo clippy --lib -- -D warnings`: clean.
- New tests cover: fallback on a missing field, fallback on broken stats, refusal on a stale-only entry, empty-history and missing-required-field errors, and the existing round-trip/caching tests (updated to generate timestamps relative to real `UtcDateTime::now()`, since staleness is checked against wall-clock time).

### DIFF
--- a/src/atp/fetch.rs
+++ b/src/atp/fetch.rs
@@ -148,11 +148,12 @@ async fn download_atp(
     Ok(metadata)
 }
 
-/// Queries `history_url` (`history.json`) and returns the metadata for the
-/// most recent run that's both complete and passes a basic sanity check
-/// (see [`MIN_SPIDERS`] and friends), as long as it isn't older than
-/// [`MAX_STALENESS`]. `history.json` entries are assumed to be listed
-/// oldest-first, as AllThePlaces does today, so we scan from the end.
+/// Queries `history_url` (`history.json`) and returns the metadata for
+/// whichever entry has the latest `start_time` among those that are both
+/// complete and pass a basic sanity check (see [`MIN_SPIDERS`] and
+/// friends), as long as it isn't older than [`MAX_STALENESS`]. Examines
+/// every entry rather than assuming `history.json` lists them in any
+/// particular order.
 async fn fetch_latest_run(client: &Client, history_url: &str) -> Result<AtpMetadata> {
     let response = client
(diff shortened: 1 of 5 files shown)


brawer (commented):
LGTM

PR #641, opened 2026-08-12
Give LineStitcher::compact() a ring-aware floor for closed loops

brawer:
## What

`LineStitcher::compact()` simplified every stored chain down to a floor of 2 coordinates, whether it was an open path or an already-closed loop. For a closed loop (`front == back`), that floor lets `SimplifyVwPreserve` strip every interior vertex, collapsing it to two identical points — a zero-area degenerate "ring" that `finish()`'s validity check then rejects and drops entirely.

## Why it mattered

Untagged, individually-closed OSM ways (e.g. a solar panel outline inside a `power=generator` multipolygon relation) get routed through `build_line`/`LineStitcher` rather than `build_ring`, relying on `GeometryBuilder`'s "promote closed stitched pieces into rings" logic to recover them (see that struct's doc comment). With enough small closed members to repeatedly exceed `LineStitcher`'s coordinate budget, every single one could collapse this way, silently erasing the whole relation's geometry.

Observed for real on OSM [relation/13606088](https://www.openstreetmap.org/relation/13606088) (808 member ways, a solar farm mapped panel-by-panel), which lost *all* its geometry despite every member way being individually well-formed — confirmed by replaying the real relation data through this exact code path before and after the fix.

## Fix

Closed chains are now simplified via a throwaway single-ring `Polygon`, so `SimplifyVwPreserve`'s ring-aware 4-point floor applies — the same trick `PolygonAssembler::compact` already uses, for the same reason. Open chains are unaffected.

## Testing

- Added `many_small_closed_loops_survive_compaction_as_valid_rings`: 40 small disjoint closed squares under a tight coordinate budget forcing repeated incremental `compact()`. Verified it fails against the old `compact()` (`finish()` returns `None`, reproducing the bug) and passes with the fix.
- All 85 existing `geometry::` tests still pass.
- `cargo clippy --all-targets` and `cargo fmt --check` are clean.

Fixes #635.

### DIFF
--- a/src/geometry/line_stitcher.rs
+++ b/src/geometry/line_stitcher.rs
@@ -4,7 +4,7 @@
 use std::collections::{HashMap, HashSet, VecDeque};
 
 use geo::algorithm::validation::Validation;
-use geo::{Coord, Geometry, LineString, MultiLineString, SimplifyVwPreserve};
+use geo::{Coord, Geometry, LineString, MultiLineString, Polygon, SimplifyVwPreserve};
 
 use super::{align_to_reference_x, centroid_x, cut_at_crossings, find_crossings};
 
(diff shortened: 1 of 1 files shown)


brawer (commented):
LGTM

PR #642, opened 2026-08-12
Assemble and embed a CycloneDX provenance BOM into conflated.parquet

brawer:
## Summary
- Rename `PbfHeader` to `OsmMetadata`, for consistency with `AtpMetadata` -- the two now sit side by side in `crate::provenance`.
- Add `pipeline::osm::read_header(path)`: a standalone entry point for `OsmMetadata` that doesn't go through `BlobReader::open()`. Every `.osm.pbf` file starts with exactly one `OSMHeader` blob, so this reads just that first blob (a few hundred bytes) and returns, instead of scanning the whole file's blob-header table the way `BlobReader::open()` has to (needed there to build the node/way/relation partition index, not needed just for the header).
- Add `atp::read_cached_metadata(workdir)`: a synchronous accessor reading `AtpMetadata` back from `alltheplaces.meta.json`. Unlike `fetch_atp`/`read_meta_json`, this doesn't need a tokio runtime -- by the time it's useful, the file is already on disk and we're outside any async context.
- Add `crate::provenance::build(workdir)`: reads both metadata structs back from `workdir` and assembles a CycloneDX 1.7 JSON document -- `metadata.component` for the pipeline's own version, one `type: "data"` component per input source -- following this repo's existing "data"-component shape from `scripts/sbom/pipeline.jq`. Depends only on `AtpMetadata`/`OsmMetadata`, not on `PlaceIndex`/`FeatureStore` or any of the indexing built on top of them, so it isn't coupled to the OSM indexing revamp in progress.
- `conflate()`'s writer embeds the serialized document as the output file's `cyclonedx.bom` Parquet key-value metadata.

`OsmMetadata`/`read_header`/`PLANET_PBF_FILENAME` and `AtpMetadata`/`read_cached_metadata` are re-exported `pub(crate)` narrowly from `pipeline`/`atp` (rather than making the whole `osm`/`fetch` modules `pub(crate)`), so the rest of their APIs (`BlobReader`, `Node`/`Way`/`Relation`, `fetch_atp`, ...) stay exactly as private as before.

## Why
Closes #638. Builds on #634 (OSM) and #640 (AllThePlaces), which parsed and logged this metadata but didn't do anything with it yet.

## Testing
- `cargo test`: 227 lib tests + 4 integration tests pass.
- `cargo clippy --all-targets -- -D warnings` and `cargo fmt --check`: clean.
- Manually verified end-to-end against the real test fixtures (not just unit tests): ran the actual binary against a workdir seeded with `tests/test_data/{alltheplaces.zip,alltheplaces.meta.json,zugerland.osm.pbf}`, then inspected the resulting `conflated.parquet`'s embedded metadata directly:
  ```json
  {"bomFormat":"CycloneDX","specVersion":"1.7","metadata":{"component":{"name":"osm-diffs","type":"application","version":"0.6.10"}},"components":[{"type":"data","name":"alltheplaces","version":"2026-01-01-00-00-00","properties":[{"name":"alltheplaces:run_id","value":"2026-01-01-00-00-00"}, ...]},{"type":"data","name":"openstreetmap-planet","version":"2026-01-27T08:11:02Z","properties":[{"name":"osm:replication_timestamp","value":"2026-01-27T08:11:02Z"},{"name":"osm:writing_program","value":"osmx"}]}]}
  ```

### DIFF
--- a/src/pipeline/conflate/writer.rs
+++ b/src/pipeline/conflate/writer.rs
@@ -75,7 +75,16 @@ pub struct ParquetRow {
 
 /// Key under which the CycloneDX provenance BOM (see `crate::provenance`)
 /// is stored in this file's Parquet key-value metadata.
-const PROVENANCE_KEY: &str = "cyclonedx.bom";
+///
+/// Neither CycloneDX nor Parquet defines a convention for this: CycloneDX
+/// is transport-agnostic (a BOM is normally a sibling file, an OCI
+/// artifact, or an in-toto/SLSA attestation, none of which map to a
+/// Parquet key), and Parquet's `key_value_metadata` is just a flat list
+/// of arbitrary strings that tools namespace themselves (e.g. GeoParquet
+/// uses `geo`, GDAL `gdal:schema`, Spark
+/// `org.apache.spark.sql.parquet.row.metadata`, Arrow `ARROW:schema`).
+/// `org.cyclonedx.bom` follows that same reverse-DNS-style namespacing.
+const PROVENANCE_KEY: &str = "org.cyclonedx.bom";
 
 impl ParquetWriter {
     /// `provenance_bom` is the CycloneDX document from `crate::provenance`,
--- a/src/pipeline/conflate/writer.rs
+++ b/src/pipeline/conflate/writer.rs
@@ -11,7 +11,7 @@ use deepsize::DeepSizeOf;
 use parquet::{
     arrow::{ArrowWriter, arrow_writer::ArrowWriterOptions},
     basic::{Compression, ZstdLevel},
-    file::properties::WriterProperties,
+    file::{metadata::KeyValue, properties::WriterProperties},
     schema::types::ColumnPath,
 };
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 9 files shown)


src/pipeline/conflate/writer.rs
```diff
@@ -73,14 +73,29 @@ pub struct ParquetRow {
     atp_shape_wkb: Vec<u8>,
 }
 
+/// Key under which the CycloneDX provenance BOM (see `crate::provenance`)
+/// is stored in this file's Parquet key-value metadata.
+const PROVENANCE_KEY: &str = "cyclonedx.bom";
```
brawer: Make this "org.cyclonedx.bom". Also, add a comment to summarize the rationale (see below). We'll also need to document this on behalf of downstream consumers, who won't look into the code — but that will be a separate, much larger task.

So I asked Claude the following question: When embedding a CycloneDX BOM as file metadata into a Parquet file, is there any convention on what metadata key to use? This is for describing how the Parquet file was built (in particular, the provenance of its inputs, plus the software version that produced the file).

Claude answered the following:

No, there isn't an established or CycloneDX-published convention for this. A few relevant data points:

CycloneDX itself is transport/format-agnostic. The spec defines the BOM document format but says nothing about how to embed it inside other file formats — it's normally shipped as a sibling file (bom.json / *.cdx.json), attached to an OCI artifact (media type application/vnd.cyclonedx+json), or wrapped in an in-toto/SLSA attestation using the predicate type https://cyclonedx.org/bom. None of that maps to a Parquet key.
Parquet's FileMetaData.key_value_metadata is just a flat list of arbitrary string key/value pairs, and in practice tools namespace their own keys rather than following any registry (e.g. GeoParquet uses geo, GDAL uses gdal:schema, Spark uses org.apache.spark.sql.parquet.row.metadata, Arrow uses ARROW:schema).

Since nothing formal exists, the practical options are:

A short, unambiguous key like cyclonedx or cdx, or a reverse-DNS-style key like org.cyclonedx.bom (matching the namespacing style Spark/Arrow use), with the raw CycloneDX JSON as the value.
If you want alignment with the attestation ecosystem instead, use the CycloneDX predicate-type URI https://cyclonedx.org/bom as the key — that's at least a string CycloneDX itself defines, even though it wasn't designed for this purpose.

I'd lean toward org.cyclonedx.bom — it reads clearly, avoids collisions with other tooling, and follows the namespacing pattern already used by GDAL/Spark/Arrow for Parquet metadata keys. Just document your choice in a README next to the file, since a downstream reader has no way to discover it otherwise.

src/provenance.rs
```diff
@@ -0,0 +1,175 @@
...
+    let osm_planet = workdir.join(pipeline::PLANET_PBF_FILENAME);
+    let osm_metadata =
+        pipeline::read_header(&osm_planet).context("could not read OpenStreetMap provenance")?;
+
+    Ok(json!({
+        "bomFormat": "CycloneDX",
+        "specVersion": "1.7",
+        "metadata": {
+            "component": {
+                "type": "application",
```
brawer: No, this is wrong. We're actually trying to describe the provenance of a data file here (what sources and which software was used to build our output file), whereas `metadata.component.type=application` would be appropriate for a piece of software. I'll fix this manually in a follow-up change. For now, just change "application" to "data", and add a TODO comment before the `json!` macro, stating that this BOM is not correct yet.

src/pipeline/conflate/writer.rs
brawer: Done in 01aba24: key is now `org.cyclonedx.bom`, with a comment on `PROVENANCE_KEY` summarizing why (no CycloneDX/Parquet convention exists; matches the reverse-DNS namespacing GDAL/Spark/Arrow already use for their own Parquet keys). Agreed that documenting this for downstream consumers is a separate, larger task -- not attempting that here.

src/provenance.rs
brawer: Done in 01aba24: changed to `type: "data"`, and added a TODO right above the `json!` noting that `metadata.component` isn't modeled correctly yet overall, pending your follow-up.

brawer (commented):
LGTM

Issue #645, opened 2026-08-12
Write docs/CONFLATED_OUTPUT.md describing the conflated.parquet schema


brawer:
## Context

#644 fixes `crate::provenance` to correctly model the CycloneDX 1.7 data
BOM embedded in `conflated.parquet`. Its `metadata.component` (the
output-file component) includes a `documentation`-type external
reference pointing at `docs/CONFLATED_OUTPUT.md` -- which doesn't exist
yet. The link is intentionally left broken until this is written (see
the `externalReferences` entry in `src/provenance.rs`, and the code
comment there referencing this issue).

## What to do

Write `docs/CONFLATED_OUTPUT.md`, describing the schema of
`conflated.parquet` for downstream consumers who only have the output
file (and, now, the embedded BOM) -- not this repo's source. At minimum:

- The Arrow/Parquet schema itself: columns, types, nesting (see
  `src/pipeline/conflate/writer.rs`'s `schema` module for the ground
  truth).
- What `atp.*` / `osm.*` fields mean and how they relate to the
  original AllThePlaces/OpenStreetMap data.
- How to read the embedded `org.cyclonedx.bom` key-value metadata (the
  provenance BOM from #644), and what it means.

## Priority

Not urgent -- filed per #644's own instructions, to close the loop on a
link that's currently broken by design. Best done once the Parquet
schema itself has stabilized.

PR #647, opened 2026-08-12
Fix crate::provenance to correctly model the CycloneDX 1.7 data BOM

brawer:
## Summary
- `metadata.component` now describes `conflated.parquet` itself (`type: "data"`, `mime-type: "application/vnd.apache.parquet"`, `version` = this run's timestamp) instead of conflating the *tool* with the *data file* -- fixes the TODO left in #642's review.
- `metadata.tools.components[0]` describes the `osm-diffs` tool instead (`type: "application"`, version, `purl`, `vcs`/`release-notes` external references). Also resolves #588: the tool's own version now lives in the right place.
- `components[]` drops #642's `properties`-array approach for properly modeled fields: `version` carries each source's own timestamp (`AtpMetadata.start_time` / `OsmMetadata.replication_timestamp` -- not `run_id`, confirmed redundant with `start_time`), `data`/`contents`/`externalReferences` carry the rest. Run-health stats and values constant across every run are dropped entirely -- not provenance, just noise logs already cover.
- New `formulation[]` ties the tool, both inputs, and the output together into one workflow.
- `serialNumber` and `formulation[].workflows[].uid` share one UUID (document identity vs. execution identity -- different concerns in general, but this pipeline never runs the same workflow twice, so they're always 1:1).
- Added `uuid` as a direct dependency (`v4` feature only) -- already present transitively via `librqbit`, so nothing new in the supply chain.

## Why
Closes #644, **except** SHA-256 hashing of both downloaded input files, split out into #646 (blocking #644, per discussion -- computed from bytes we don't capture today, unlike everything else here). `src/provenance.rs` has two `// TODO(#646)` comments marking where the resulting `hashes` entries belong.

Also files, per #644's own instructions: #645 (write `docs/CONFLATED_OUTPUT.md`, linked from `metadata.component`'s `documentation` external reference -- a broken link until that's written).

## Testing
- `cargo test`: 230 lib tests + 4 integration tests pass.
- One-time `cyclonedx-cli validate --input-version v1_7 --fail-on-errors` against a BOM generated from the real test fixtures: **passes**. This caught a real gap in #644's example shape -- `workflow.uid` is required by the actual CycloneDX 1.7 schema but wasn't in the example (added, with a comment explaining why).

### DIFF
--- a/src/provenance.rs
+++ b/src/provenance.rs
@@ -146,6 +146,10 @@ fn atp_component(atp: &AtpMetadata) -> Value {
         "name": "alltheplaces",
         "version": &start_time,
         "supplier": supplier(),
+        // Using AllThePlaces data in OpenStreetMap (this pipeline's
+        // purpose) was discussed with -- and not objected to by -- the
+        // OSM Foundation's Licensing Working Group:
+        // https://osmfoundation.org/wiki/Licensing_Working_Group/Minutes/2023-08-14#Ticket%232023081110000064_%E2%80%94_First_party_websites_as_sources
         "licenses": license("CC0-1.0"),
         // TODO(#646): checksum is a placeholder until we compute a real
         // SHA-256 hash of the downloaded zip; the purl is invalid until
--- a/src/provenance.rs
+++ b/src/provenance.rs
@@ -26,6 +26,18 @@ use uuid::Uuid;
 /// `metadata.tools.components[0]`'s external references and purl.
 const REPO_URL: &str = "https://github.com/alltheplaces/osm-diffs";
 
+/// Supplier declared for this BOM and both input components. Copied
+/// verbatim from `scripts/sbom/merge.jq`'s `metadata.supplier` (the
+/// container-image SBOM) rather than kept in sync programmatically --
+/// the project name won't change, and if it ever does, a grep finds
+/// both places.
+fn supplier() -> Value {
+    json!({
+        "name": "All The Places",
+        "url": ["https://github.com/alltheplaces/"]
+    })
+}
+
 /// Builds the CycloneDX provenance document for `conflated.parquet`,
 /// from whatever `import_atp`/`import_osm` already left behind in
 /// `workdir`. `pipeline_run_id` becomes
@@ -54,6 +66,7 @@ pub fn build_bom_for_conflated_parquet(workdir: &Path, pipeline_run_id: &str) ->
         "version": 1,
         "metadata": {
             "timestamp": run_timestamp,
+            "supplier": supplier(),
             "tools": {
                 "components": [tool_component()],
             },
@@ -63,6 +76,16 @@ pub fn build_bom_for_conflated_parquet(workdir: &Path, pipeline_run_id: &str) ->
             atp_component(&atp_metadata),
             osm_component(&osm_metadata),
         ],
+        // Declares conflated.parquet's two inputs as *its* dependencies,
+        // so the dependency graph has a single root (conflated.parquet)
+        // instead of three unlinked ones -- an NTIA-minimum-elements
+        // validator flags components no other component depends on as
+        // extra roots. Same pattern as scripts/sbom/pipeline.jq uses for
+        // the container-image SBOM.
+        "dependencies": [{
+            "ref": "conflated.parquet",
+            "dependsOn": ["alltheplaces.zip", pipeline::PLANET_PBF_FILENAME],
+        }],
         "formulation": [formulation(pipeline_run_id)],
     }))
 }
@@ -110,6 +133,7 @@ fn atp_component(atp: &AtpMetadata) -> Value {
         "type": "data",
         "name": "alltheplaces",
         "version": format_rfc3339(atp.start_time),
+        "supplier": supplier(),
         "data": [{"type": "dataset", "contents": {"url": atp.output_url}}],
         "externalReferences": [
             {"type": "distribution", "url": atp.output_url},
@@ -120,6 +144,7 @@ fn atp_component(atp: &AtpMetadata) -> Value {
 }
 
 fn osm_component(osm: &OsmMetadata) -> Value {
+    let replication_timestamp = format_rfc3339(osm.replication_timestamp);
     json!({
         // Reference PLANET_PBF_FILENAME rather than a literal, so this
         // can't drift from the file's actual local name (see #648 for a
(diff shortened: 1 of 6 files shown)


src/provenance.rs
```diff
@@ -1,10 +1,10 @@
 //! Assembles this pipeline's provenance -- which input data, and which
 //! version of the pipeline itself, produced a given output file -- as a
-//! JSON document modeled on a CycloneDX 1.7 Bill of Materials, but
-//! describing data lineage rather than code dependencies (component
-//! `type: "data"`, one per input source, following the same shape this
-//! repo already uses for non-crate "data" components in the build-time
-//! code SBOM, see `scripts/sbom/pipeline.jq`). See issue #638.
+//! CycloneDX 1.7 Bill of Materials describing data lineage rather than
```
brawer: Readers won't care about the task description of #644, this is too noisy as long-term documentation. Instead, just say something like the following. Feel free to rewrite my blurb, so that readers unfamilar with BOMs or supply chains understand why we do this.

Assembles the provenance of this pipeline's public output — a machine-readable document that describes which version of this pipeline, using exactly what input, at which time, produced an output file. In Parquet, we embed this document into the file metadata. We use the industry-standard CycloneDX JSON format.

src/provenance.rs
```diff
@@ -31,65 +36,123 @@ pub fn build(workdir: &Path) -> Result<Value> {
     let osm_metadata =
         pipeline::read_header(&osm_planet).context("could not read OpenStreetMap provenance")?;
 
-    // TODO: metadata.component isn't modeled correctly yet. We're
-    // describing the provenance of a *data* file here (which sources,
-    // and which software, built our output), not shipping a BOM for a
-    // piece of software -- so `type: "application"` is wrong for this
-    // component. To be fixed in a follow-up.
+    let run_timestamp = format_rfc3339(UtcDateTime::now());
+    // One UUID, reused for both: `serialNumber` identifies this BOM
```
brawer: Actually, we only need this UUID for `serialNumber`, see below.

src/provenance.rs
brawer: Something like "build_bom_for_conflated_parquet()" would be more descriptive; "build" could mean anything. Eventually, we'll likely have more outputs.

src/provenance.rs
```diff
@@ -31,65 +36,123 @@ pub fn build(workdir: &Path) -> Result<Value> {
...
-        "value": replication_timestamp,
-    })];
-    if let Some(source) = &osm.source {
-        properties.push(json!({"name": "osm:source", "value": source}));
-    }
-    if let Some(writing_program) = &osm.writing_program {
-        properties.push(json!({"name": "osm:writing_program", "value": writing_program}));
-    }
     json!({
+        "bom-ref": "src-osm-planet",
```
brawer: "openstreetmap-planet"

src/provenance.rs
```diff
@@ -31,65 +36,123 @@ pub fn build(workdir: &Path) -> Result<Value> {
...
+/// which inputs to produce which output. `uid` is required by the
+/// CycloneDX 1.7 schema even though it's absent from #644's example
+/// shape (confirmed via `cyclonedx-cli validate` against the real
+/// schema) -- see its doc comment in `build()` for what it identifies.
+fn formulation(workflow_uid: &str) -> Value {
+    json!({
+        "bom-ref": "formula-osm-diffs-build",
+        "workflows": [{
+            "bom-ref": "workflow-osm-diffs-build",
+            "uid": workflow_uid,
```
brawer: Actually, turns out `uid` has a different semantics in CycloneDX than I thought at first. It is not necessarily an UUID, but can be any string ID that identifies an individual run of our pipeline. For example, this could be a Kubernetes run identifier, or whatever other ID is used by the system that executes the container for our pipeline. We can't compute this value ourselves here: rather, we need to take it from the outside. Please add a command-line parameter `--run_id` to main.rs (defaulting to an empty string), and wire it all the way down to here, perhaps calling it `pipeline_run_id` in places where's ambiguity between different kinds of runs. Perhaps we'll come up with a different way of identifying our pipeline runs in the long term, but for now, this will do.

src/provenance.rs
```diff
@@ -31,65 +36,123 @@ pub fn build(workdir: &Path) -> Result<Value> {
...
+        ],
+    })
+}
+
+/// `conflated.parquet` itself -- the data file this BOM is embedded
+/// into, described as data (not as the `osm-diffs` tool that built it,
+/// which is `tool_component()` instead).
+fn output_component(run_timestamp: &str) -> Value {
+    json!({
+        "bom-ref": "output-conflated",
```
brawer: "conflated.parquet" (if this passes validation), otherwise "conflated-parquet".

src/provenance.rs
```diff
@@ -31,65 +36,123 @@ pub fn build(workdir: &Path) -> Result<Value> {
...
+        "version": format_rfc3339(osm.replication_timestamp),
+        "data": [{"type": "dataset"}],
+        // TODO(#646): add a "hashes" entry (SHA-256 of the downloaded
+        // .osm.pbf) once we compute one. Not done today.
+    })
+}
+
+/// Ties `tool_component()`, the two input `*_component()`s, and
+/// `output_component()` together into one workflow: which tool consumed
+/// which inputs to produce which output. `uid` is required by the
```
brawer: Stop after "output." Remove the second sentence (after `uid`).

src/provenance.rs
brawer: Done in 5a9953d, using your blurb (lightly trimmed/adjusted): rewrote the module doc comment for someone unfamiliar with BOMs/supply chains, dropped the #644/#638/pipeline.jq task-history references. Kept one paragraph on why metadata is read straight back from workdir rather than threaded through import_atp/import_osm -- that's a load-bearing design decision, not task noise, so I thought worth keeping as long-term documentation.

src/provenance.rs
brawer: Done in 5a9953d: only serialNumber gets a minted UUID now. workflow_uid comes from the new --run_id CLI flag instead (see the reply on the uid-semantics comment below).

src/provenance.rs
brawer: Done in 5a9953d: renamed to build_bom_for_conflated_parquet().

src/provenance.rs
brawer: Done in 5a9953d: bom-ref is now src-openstreetmap-planet, matching its name field (the same way src-alltheplaces already matched "alltheplaces") -- that was the only one of the four bom-refs that didn't.

src/provenance.rs
brawer: Done in 5a9953d: added --run_id to `osm-diffs run` (default: empty string), threaded through as pipeline_run_id in run_pipeline -> conflate -> write_conflated -> build_bom_for_conflated_parquet, which now uses it verbatim as formulation[].workflows[].uid instead of a minted UUID. Manually verified with --run_id gh-actions-run-12345 that it lands in the embedded BOM correctly.

src/provenance.rs
brawer: Confirmed via cyclonedx-cli validate --input-version v1_7 --fail-on-errors against a real generated BOM: "conflated.parquet" (with the dot) validates fine. Kept as-is.

src/provenance.rs
brawer: Done in 5a9953d.

src/provenance.rs
brawer: Good to know. So, in that case, can we change "output-conflated" to "conflated.parquet"?

PR #649, opened 2026-08-13
Rename PLANET_PBF_FILENAME to match upstream's planet-latest.osm.pbf

brawer:
## Summary
Renames `PLANET_PBF_FILENAME` from `osm-planet.pbf` (a name we made up locally) to `planet-latest.osm.pbf`, matching upstream's own name for this file (`OSM_TORRENT_URL` in `src/pipeline/osm/fetch.rs`).

`crate::provenance`'s `osm_component()` already referenced the constant rather than a literal, so it picks up the rename automatically -- confirmed by re-running the pipeline end-to-end and checking the embedded BOM. Only two places needed an actual edit: the constant definition, and `tests/integration_test.rs`'s fixture symlink (a separate binary/test target, can't reach the `pub(crate)` constant).

## Why
Closes #648.

## Testing
- `cargo test`: 231 lib + 4 integration tests pass.
- Manually verified end-to-end: ran the binary against a workdir with the input renamed to `planet-latest.osm.pbf`, confirmed the embedded BOM's OSM component's `bom-ref`/`name` both read `planet-latest.osm.pbf`.

### DIFF
--- a/src/pipeline/osm/mod.rs
+++ b/src/pipeline/osm/mod.rs
@@ -249,8 +249,10 @@ fn build_relation_parents<R: Read + Seek + Send>(
 
 /// Filename, within `workdir`, that `fetch::fetch_planet` downloads the
 /// OSM planet PBF to. Shared with `crate::provenance`, which needs to
-/// find it again (via `read_header`) without re-fetching.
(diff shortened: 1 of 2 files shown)


brawer (commented):
LGTM

PR #650, opened 2026-08-13
Compute and embed the AllThePlaces zip's SHA-256 hash (#646, ATP half)

brawer:
## Summary
Addresses the **AllThePlaces half** of #646 -- the OSM planet PBF's hash stays a `checksum=sha256:TODO` placeholder for now, left for separate follow-up work (needs new plumbing: OSM has no metadata sidecar file the way ATP already does, and hashing a multi-GB file isn't cheap-to-redo-on-demand the way `OsmMetadata::read_header()` is).

- `AtpMetadata` gets a new `sha256: Option<String>` field: `None` for a candidate run fresh out of `history.json` (nothing downloaded yet to hash), `Some(hex)` once `fetch_atp` has actually downloaded and hashed the file. Computed via `aws_lc_rs::digest::Context`, streamed alongside the existing chunk-by-chunk download/write loop in `download_atp()` -- the same crypto library this crate already uses for TLS, not a second, separate hashing implementation.
- `provenance::atp_component()` now populates a real `hashes` entry and a valid `checksum=sha256:<hash>` purl qualifier. Errors out if `sha256` is ever missing at BOM-assembly time (rather than a silent fallback) -- `import_atp` always runs before `conflate`, so this shouldn't happen in practice; if it does, that's worth surfacing, not papering over.
- `tests/test_data/alltheplaces.meta.json` (used by the real end-to-end integration test) gets the actual SHA-256 of the checked-in `alltheplaces.zip` fixture.

One thing worth knowing: this build links `aws-lc-sys`, not `aws-lc-fips-sys` -- the general-purpose backend, not AWS-LC's FIPS-140-3-validated module. Same library lineage, just not that build configuration; noted in the code comment.

## Why
Part of #646 (not closing it -- OSM's half is still open).

## Testing
- `cargo test`: 232 lib tests + 4 integration tests pass.
- New unit test computes SHA-256("abc") through the exact same `aws_lc_rs` call path `download_atp()` uses and checks it against NIST's published worked example for that message.
- Manually verified end-to-end against the real test fixtures: the embedded BOM's `alltheplaces.zip` component now carries the real SHA-256 of the fixture zip, in both `hashes` and the `purl`.


brawer (commented):
LGTM

PR #652, opened 2026-08-13
Link SECURITY.md to SUPPLY_CHAIN_SECURITY.md

brawer:
## Summary
- `docs/SECURITY.md` covers vulnerability reporting; `docs/SUPPLY_CHAIN_SECURITY.md` covers a different concern — verifying releases via provenance, SBOMs, and signed attestations.
- A reader landing on `SECURITY.md` had no path to the supply-chain doc, so add a one-line pointer at the end (kept separate from the vulnerability-reporting instructions).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #653, opened 2026-08-13
docs/SECURITY.md: use an absolute link to SUPPLY_CHAIN_SECURITY.md

brawer:
## Summary
Follow-up to #652.

The relative link added there works on the ordinary blob view (https://github.com/alltheplaces/osm-diffs/blob/main/docs/SECURITY.md) but not on GitHub's security-tab view (https://github.com/alltheplaces/osm-diffs?tab=security-ov-file), which renders SECURITY.md's content outside its normal blob context — relative links there resolve against the repo root instead of `docs/`, so `SUPPLY_CHAIN_SECURITY.md` pointed at a nonexistent top-level file.

Switched to an absolute URL, which resolves correctly in both places (matching the existing absolute link to the security tab itself, just above it).


brawer (commented):
LGTM

PR #654, opened 2026-08-13
Compute and embed the OSM planet file's SHA-256 hash (#646, OSM half)

brawer:
## Summary

The OSM half of #646 (the AllThePlaces half was #650): computes and
embeds a real SHA-256 hash of the downloaded OSM planet file
(`planet-latest.osm.pbf`), replacing `crate::provenance`'s
`checksum=sha256:TODO` placeholder for that component.

- `pipeline::osm::compute_and_persist_metadata` hashes the freshly
  downloaded planet file in one dedicated sequential read pass (fixed
  8 MiB buffered chunks, not `memmap2::Mmap`, to keep memory flat
  regardless of file size -- this file is tens of GB) and persists the
  result to a `planet-latest.osm.pbf.meta.json` sidecar in `workdir`,
  alongside the header fields we already parse. A later run against
  the same `workdir` reads this back (`read_cached_metadata`) instead
  of re-hashing.
- Unlike AllThePlaces' plain sequential HTTP download, the planet file
  arrives via BitTorrent (`librqbit`), which writes pieces to disk
  itself, out of order -- so unlike ATP's hash, this one can't be
  computed incrementally during download; it has to be a separate pass
  over the completed file.
- Both hashes go through the same `aws_lc_rs::digest` (SHA-256) call
  path already used for this pipeline's TLS -- no second hashing
  implementation.
- `crate::provenance`'s `osm_component()` now emits a real
  `hashes`/`purl` checksum for the OSM component, mirroring
  `atp_component()`'s existing pattern exactly.
- Pulled the `to_hex`/RFC3339-serde helpers `atp::fetch` had grown out
  into a new `crate::utils` module, now shared by both `AtpMetadata`
  and `OsmMetadata`, instead of duplicating them for OSM.

## Verification

Ran the full pipeline against the test fixtures, then:
- confirmed both components' embedded hashes match `shasum -a 256`
  computed independently on the same files
- validated the resulting BOM with `cyclonedx validate`
- validated the resulting BOM with FOSSA's NTIA SBOM validator
  (https://fossa.com/resources/regulatory-compliance-tools/ntia-sbom-validator/)
  -- it still passes

Closes #646.

### DIFF
--- a/src/atp/fetch.rs
+++ b/src/atp/fetch.rs
@@ -1,4 +1,5 @@
 use crate::make_download_bar;
+use crate::utils::to_hex;
 use anyhow::{Context, Ok, Result, anyhow};
 use aws_lc_rs::digest::{Context as DigestContext, SHA256};
 use futures_util::StreamExt;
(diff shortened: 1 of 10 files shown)


brawer (commented):
LGTM

Issue #655, opened 2026-08-13
Retire the old OSM pipeline; conflate against the new geometry-aware path


brawer:
Design doc / tracking issue for retiring the old OSM pipeline
(`filter.rs`/`old_assemble.rs`/`cover.rs`) and running conflation against
the new geometry-aware path (`prune.rs`/`assemble.rs`) instead. Landing as
a staged sequence of PRs (see "Staging" below); this issue tracks the
whole effort and links each PR as it lands.

## Context

OpenStreetMap is currently processed twice per pipeline run:

- **Old path** (`src/pipeline/osm/filter.rs` + `old_assemble.rs`, driven by
  `cover.rs`): scopes OSM down to whatever lies within AllThePlaces'
  geographic coverage, and writes `osm.parquet` as `Place` rows (point-only,
  no real geometry; relations are skipped entirely). This is what
  `conflate` and `suggest_edits` actually consume today.
- **New path** (`prune.rs` + `assemble.rs`): already runs on every pipeline
  invocation, but its result (`Assembly`) is thrown away
  (`src/pipeline/osm/mod.rs:63-66`, `let _assembly = ...; if false { todo!(); }`).
  It scopes OSM by tag category instead (global, not ATP-footprint-scoped),
  and — critically — computes a real, OGC-valid `geo::Geometry` for every
  kept node/way/leaf-relation/super-relation, encoded as WKB, plus an S2
  centroid and coverage cells for spatial indexing.

Goal: delete the old path, and make `conflate`/`suggest_edits` run against
the new path's output — which also finally lets `conflated.parquet` carry
real OSM geometry (today `osm_shape_wkb` in `conflated.parquet` is always a
synthetic point reconstructed from an S2 cell, via `Place::shape()`,
because `Place` never carried real geometry).

Net effect on cost is not yet known and needs measuring, not assuming.
The new path already runs unconditionally today (just discarded), and this
change removes the old path's redundant cover/filter/assemble pass, which
points toward a win. But the two paths aren't directly comparable by
scope alone: the old path's tag-category-vs-ATP-footprint scoping
difference is one axis, but the storage formats differ too — the old
path's `osm.parquet` is heavily compressed (ZSTD level 22) columnar
storage, while the new feature index is an uncompressed, mmap'd structure
designed to lean on the OS/CPU's page cache instead. Whether that nets out
faster/smaller/more memory-hungry in practice is genuinely open; treat it
[...]

PR #656, opened 2026-08-13
Add OsmFeatureIndex: spatial index over OSM Feature protos

brawer:
PR 1 of the staged sequence tracked in #655 (retiring the old OSM
pipeline in favor of the new geometry-aware path).

Adds `src/tables/feature_index.rs` (`OsmFeatureIndex`), standalone —
not wired into the pipeline yet, deliberately, so the riskiest and most
novel piece (the index's on-disk query semantics) lands and gets
reviewed/tested in isolation before anything depends on it.

Two mmap'd, uncompressed on-disk structures:

- **Feature storage**: features laid out in `centroid_s2_cell_id`
  order (for locality only — never queried directly), each addressable
  by a stable `LocalFeatureRef`.
- **Inverted index**: the actual query structure. Each feature's
  `coverage_s2_cell_id` list is exploded into `(cell_id, match_mask,
  local_index)` entries, sorted by `cell_id`. A query is a decode-free
  `partition_point` scan plus a numeric `MatchMask` check; only turning
  a `LocalFeatureRef` into a full `Feature` (`get_feature`) costs a
  protobuf decode.

`centroid_s2_cell_id` is deliberately *not* used as a query key: for
non-point geometry (lines, polygons), a feature's coverage cells can
sit anywhere in centroid sort order, so only `coverage_s2_cell_id` is a
valid containment test. Covered by
`query_uses_coverage_cells_not_centroid_order`.

## Testing
- `cargo test --lib` (240 passed) and `cargo test --test
  integration_test` (4 passed)
- `cargo clippy --all-targets -- -D warnings` clean
- `cargo fmt --check` clean

### DIFF
--- a/src/tables/feature_index.rs
+++ b/src/tables/feature_index.rs
@@ -103,7 +103,14 @@ const INVERTED_INDEX_FILE_SIGNATURE: &[u8; 8] = b"featinv0";
 /// layout for a one-field tuple struct already has the same size as its
 /// field in practice; `repr(transparent)` would only make that a
 /// documented *guarantee*, which nothing here relies on.
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+///
+/// `Ord`/`PartialOrd`/`Serialize`/`Deserialize` are derived so this can be
+/// used directly as [Pass2Item]'s `local_index` field -- pass 1 computes
+/// exactly this value (a feature's position in feature storage) before
+/// it's ever wrapped and handed to a caller; ordering it just delegates
+/// to the wrapped `u32` and carries no meaning beyond "some position",
+/// same as for any other field [Pass2Item] is sorted by.
+#[derive(Clone, Copy, Debug, Deserialize, Eq, Ord, PartialEq, PartialOrd, Serialize)]
 pub struct LocalFeatureRef(u32);
 
 /// Read-only, memory-mapped spatial index over [Feature] protos. See the
@@ -133,6 +140,7 @@ pub struct OsmFeatureIndex<'a> {
 #[derive(Clone, Debug, DeepSizeOf, Deserialize, Eq, Ord, PartialEq, PartialOrd, Serialize)]
 struct Pass1Item {
     centroid_s2_cell_id: u64,
+    /// Encoded `FeatureToIndex` proto (`FeatureToIndex::encode_to_vec()`).
     fti_bytes: Vec<u8>,
 }
 
@@ -145,7 +153,7 @@ struct Pass1Item {
 struct Pass2Item {
     coverage_cell_id: u64,
     match_mask: u32,
-    local_index: u32,
+    local_index: LocalFeatureRef,
 }
 
 impl<'a> OsmFeatureIndex<'a> {
@@ -651,7 +659,7 @@ impl InvertedIndexWriter {
         })
     }
 
-    fn write(&mut self, cell_id: u64, match_mask: u32, local_index: u32) -> Result<()> {
+    fn write(&mut self, cell_id: u64, match_mask: u32, local_index: LocalFeatureRef) -> Result<()> {
         if self.entries_count > 0 && cell_id < self.last_cell_id {
             anyhow::bail!(
                 "coverage cell ids must be written in non-decreasing order, but {} < {}",
@@ -663,7 +671,7 @@ impl InvertedIndexWriter {
         self.writer.write_all(&cell_id.to_le_bytes())?;
         self.last_cell_id = cell_id;
 
-        let packed = ((match_mask as u64) << 32) | (local_index as u64);
+        let packed = ((match_mask as u64) << 32) | (local_index.0 as u64);
         self.packed_writer.write_all(&packed.to_le_bytes())?;
 
         self.entries_count += 1;
@@ -721,7 +729,7 @@ impl Iterator for Pass2Reader {
             Ok(()) => Some(Pass2Item {
                 coverage_cell_id: u64::from_le_bytes(buf[0..8].try_into().unwrap()),
                 match_mask: u32::from_le_bytes(buf[8..12].try_into().unwrap()),
-                local_index: u32::from_le_bytes(buf[12..16].try_into().unwrap()),
+                local_index: LocalFeatureRef(u32::from_le_bytes(buf[12..16].try_into().unwrap())),
             }),
             Err(_) => None,
         }
--- a/src/tables/feature_index.rs
+++ b/src/tables/feature_index.rs
@@ -33,13 +33,16 @@
 //! byte 8..16:  entry count, u64 little-endian
 //! byte 16..24: offset of the `id` array, u64 little-endian
 //! byte 24..32: offset of the `starts` array, u64 little-endian
-//! byte 32..40: offset of the `blob` region, u64 little-endian
+//! byte 32..40: offset of the `blobs` region, u64 little-endian
 //! byte 40..64: reserved, zero-filled
 //!
 //! id array:     `entry count` OSM feature ids, u64 little-endian each
-//! starts array: `entry count + 1` byte offsets into `blob`, u64
-//!               little-endian each
-//! blob region:  concatenated `Feature.encode_to_vec()` bytes
+//! starts array: `entry count + 1` byte offsets into `blobs`, u64
+//!               little-endian each -- the last entry is a sentinel
+//!               (the total size of `blobs`), so every feature's byte
+//!               range is `starts[i]..starts[i+1]` without special-casing
+//!               the final entry
+//! blobs region: concatenated `Feature.encode_to_vec()` bytes
 //! ```
 //!
 //! Inverted index (`out` with an added `.inverted` extension):
@@ -47,26 +50,32 @@
 //! ```text
 //! byte 0..8:   magic "featinv0"
 //! byte 8..16:  entry count, u64 little-endian
-//! byte 16..24: offset of the `coverage_cell_id` array, u64 little-endian
+//! byte 16..24: offset of the `coverage_cell_ids` array, u64 little-endian
 //! byte 24..32: offset of the `packed` array, u64 little-endian
 //! byte 32..64: reserved, zero-filled
 //!
-//! coverage_cell_id array: `entry count` S2 cell ids, u64 little-endian
-//!                         each, sorted ascending (duplicates allowed)
-//! packed array:           `entry count` values, u64 little-endian each,
-//!                         aligned 1:1 with coverage_cell_id: high 32
-//!                         bits are the feature's MatchMask, low 32 bits
-//!                         are its LocalFeatureRef (position in feature
-//!                         storage)
+//! coverage_cell_ids array: `entry count` S2 cell ids, u64 little-endian
+//!                          each, sorted ascending (duplicates allowed)
+//! packed array:            `entry count` values, u64 little-endian each,
+//!                          aligned 1:1 with coverage_cell_ids: high 32
+//!                          bits are the feature's MatchMask, low 32 bits
+//!                          are its LocalFeatureRef (position in feature
+//!                          storage)
 //! ```
 
 use crate::matchers::MatchMask;
+use crate::pipeline::EXTERNAL_SORT_CHUNK_BYTES;
 use crate::tables::{Feature, FeatureToIndex};
 use anyhow::{Context, Result};
-use ext_sort::{ExternalSorter, ExternalSorterBuilder, buffer::LimitedBufferBuilder};
+use deepsize::DeepSizeOf;
+use ext_sort::{
+    ExternalSorter, ExternalSorterBuilder,
+    buffer::{LimitedBufferBuilder, mem::MemoryLimitedBufferBuilder},
+};
 use memmap2::Mmap;
 use prost::Message;
 use s2::cellid::CellID;
+use serde::{Deserialize, Serialize};
 use std::{
     fs::{File, remove_file, rename},
     io::{BufReader, BufWriter, Read, Seek, SeekFrom, Write},
@@ -79,13 +88,21 @@ use std::{
 /// Size of each file's header, in bytes: see the "File format" section above.
 const HEADER_SIZE: usize = 8 * 8;
 
-const STORAGE_FILE_SIGNATURE: &[u8; 8] = b"featstg0";
-const INVERTED_FILE_SIGNATURE: &[u8; 8] = b"featinv0";
+const FEATURES_FILE_SIGNATURE: &[u8; 8] = b"featstg0";
+const INVERTED_INDEX_FILE_SIGNATURE: &[u8; 8] = b"featinv0";
 
 /// An opaque reference to a feature's position in an [OsmFeatureIndex]'s
 /// feature storage. Cheap to copy and hold onto (it's just an array
 /// index); not meaningful outside the [OsmFeatureIndex] that produced it,
 /// and not an OSM id -- see [OsmFeatureIndex::feature_id].
+///
+/// Not `#[repr(transparent)]`: that attribute buys ABI/layout guarantees
+/// for FFI or unsafe transmutation between `LocalFeatureRef` and `u32` --
+/// neither applies here, this stays entirely inside pure Rust code, never
+/// crosses an FFI boundary, and is never transmuted. Rust's default
+/// layout for a one-field tuple struct already has the same size as its
+/// field in practice; `repr(transparent)` would only make that a
+/// documented *guarantee*, which nothing here relies on.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub struct LocalFeatureRef(u32);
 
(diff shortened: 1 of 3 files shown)


src/tables/feature_index.rs
```diff
@@ -0,0 +1,790 @@
...
+
+    inverted_file: File,
+    _inverted_mmap: Mmap,
+    coverage_cell_id: &'a [u64],
+    packed: &'a [u64],
+}
+
+impl<'a> OsmFeatureIndex<'a> {
+    /// Path of the inverted-index file that accompanies `out`/`path`.
+    fn inverted_path(path: &Path) -> PathBuf {
```
brawer: Call it "inverted_index_path", otherwise it sounds like we're inverting a path, not like we're returning the path to an inverted index.

src/tables/feature_index.rs
```diff
@@ -0,0 +1,790 @@
...
+    ///    entry in each feature's `coverage_s2_cell_id` -- to a temporary
+    ///    file for pass 2.
+    /// 2. External-sort the spooled tuples by `coverage_cell_id`, and
+    ///    write the inverted-index arrays.
+    pub fn create(
+        fti: impl Iterator<Item = FeatureToIndex>,
+        workdir: &Path,
+        out: &Path,
+    ) -> Result<OsmFeatureIndex<'a>> {
+        let inverted_path = Self::inverted_path(out);
```
brawer: Likewise: inverted_index_path = Self::inverted_index_path(out)

src/tables/feature_index.rs
```diff
@@ -0,0 +1,790 @@
...
+            let mut p = PathBuf::from(out);
+            p.add_extension("pass2.tmp");
+            p
+        };
+
+        // Pass 1.
+        let sorter1: ExternalSorter<(u64, Vec<u8>), std::io::Error, LimitedBufferBuilder> =
+            ExternalSorterBuilder::new()
+                .with_tmp_dir(workdir)
+                .with_buffer(LimitedBufferBuilder::new(
```
brawer: Let's use a MemoryLimitedBufferBuilder here. This restricts the total size of chunks to a given number of bytes, not a fixed number of entries. The Vec<u8> can be quite uneven in size: From a single point with just one or two tag, to a feature with very large geometry. So it's better to limit the sorting chunks by total number of bytes, not by number of entries. For the maximal chunk size, 512 MB may be reasonable. Add this constant not here, but in src/pipeline/mod.rs.

Also, please file a GitHub issue to use this new constant across the pipeline, so that sorting chunks all have the same size. This has grown over time, there's no good reason why the chunks for external sorting should be so uneven in size. But don't change other code now, let's do this as clean-up on a rainy day.

src/tables/feature_index.rs
```diff
@@ -0,0 +1,790 @@
...
+                .build()?;
+        let sorted1 = sorter1.sort(fti.map(|f| {
+            let key = f.centroid_s2_cell_id;
+            std::io::Result::Ok((key, f.encode_to_vec()))
+        }))?;
+
+        let mut storage_writer = StorageWriter::create(out)?;
+        let mut pass2_writer =
+            BufWriter::with_capacity(64 * 1024, File::create(&pass2_spool_path)?);
+        for (local_index, item) in (0_u32..).zip(sorted1) {
```
brawer: Why not enumerate() — is this to get the local index as u32 instead of usize? An explicit cast might be clearer.

src/tables/feature_index.rs
```diff
@@ -0,0 +1,790 @@
...
+                pass2_writer.write_all(&local_index.to_le_bytes())?;
+            }
+        }
+        storage_writer.close()?;
+        pass2_writer.flush()?;
+        drop(pass2_writer);
+
+        // Pass 2.
+        let pass2_records = Pass2Reader::open(&pass2_spool_path)?;
+        let sorter2: ExternalSorter<(u64, u32, u32), std::io::Error, LimitedBufferBuilder> =
```
brawer: Here, LimitedBufferBuilder is fine since the item size is constant (8 + 4 + 4 = 16), so we don't need to compute the byte size of every item being sorted. But let's compute the number of items (the first parameter passed to LimitedBufferBuilder::new()) from the same constant that was passed to the MemoryLimitedBufferBuilder in pass 1. To make this read cleanly, define a local type for the items being sorted here, and use the Rust sizeof function as divisor.

src/tables/feature_index.rs
```diff
@@ -0,0 +1,790 @@
...
+            &storage_mmap,
+            usize::try_from(u64::from_le(header[2]))?,
+            entries_count,
+            path,
+            "id",
+        )?;
+        let starts = read_u64_array(
+            &storage_mmap,
+            usize::try_from(u64::from_le(header[3]))?,
+            entries_count + 1,
```
brawer, line 304: Add a comment, explaining that the + 1 is for the final sentinel.

src/tables/feature_index.rs
```diff
@@ -0,0 +1,790 @@
...
+                unsafe {
+                    let ptr = storage_mmap.as_ptr().add(blob_offset);
+                    std::slice::from_raw_parts(ptr, storage_mmap.len() - blob_offset)
+                }
+            } else {
+                anyhow::bail!("misplaced blob in OsmFeatureIndex: {}", path.display());
+            }
+        };
+
+        let inverted_path = Self::inverted_path(path);
```
brawer: inverted_index_path

src/tables/feature_index.rs
```diff
@@ -0,0 +1,790 @@
...
+        let inverted_file = File::open(&inverted_path).with_context(|| {
```
brawer: inverted_index_file (we're not inverting a file)

src/tables/feature_index.rs
```diff
@@ -0,0 +1,790 @@
...
+        Ok(OsmFeatureIndex {
+            storage_file,
+            _storage_mmap: storage_mmap,
+            entries_count,
+            id,
+            starts,
+            blob,
+
+            inverted_file,
+            _inverted_mmap: inverted_mmap,
```
brawer: inverted_index_mmap — we don't invert an mmap, we mmap an inverted index

src/tables/feature_index.rs
```diff
@@ -0,0 +1,790 @@
...
+        }
+
+        // SAFETY: inverted_mmap.len() checked above; offset 0 is aligned for u64.
+        let inverted_header = unsafe {
+            let ptr = inverted_mmap.as_ptr() as *const u64;
+            std::slice::from_raw_parts(ptr, HEADER_SIZE / size_of::<u64>())
+        };
+        let inverted_entries_count = usize::try_from(u64::from_le(inverted_header[1]))?;
+
+        let coverage_cell_id = read_u64_array(
```
brawer: Call it "coverage_cell_ids", since it contains (very many) ids, not just one

src/tables/feature_index.rs
```diff
@@ -0,0 +1,790 @@
...
+            &inverted_mmap,
+            usize::try_from(u64::from_le(inverted_header[3]))?,
+            inverted_entries_count,
+            &inverted_path,
+            "packed",
+        )?;
+
+        Ok(OsmFeatureIndex {
+            storage_file,
+            _storage_mmap: storage_mmap,
```
brawer: features_mmap might be a better name (across the entire file, of course). It's quite obvious that we store something, so the name should focus on what is being stored.

src/tables/feature_index.rs
```diff
@@ -0,0 +1,790 @@
...
+            entries_count,
+            id,
+            starts,
+            blob,
```
brawer: "blobs" — We store many blobs, not just one.

src/tables/feature_index.rs
```diff
@@ -0,0 +1,790 @@
...
+
+            inverted_file,
```
brawer: inverted_index_file

src/tables/feature_index.rs
```diff
@@ -0,0 +1,790 @@
...
+        &self,
+        range: RangeInclusive<CellID>,
+        query_mask: MatchMask,
+    ) -> impl Iterator<Item = LocalFeatureRef> + '_ {
+        let lo = range.start().0;
+        let hi = range.end().0;
+        let start = self
+            .coverage_cell_id
+            .partition_point(|&k| u64::from_le(k) < lo);
+        let end = self
```
brawer: We don't need to call `partition_point()` a second time here, doing a second O(log n) search. (For the full planet, n is half a billion features, and this is the core inner loop of our conflation pipeline, so this is the one place that really should be performant). Instead of finding the end by calling `partition_point()` a second time, just do a linear scan beginning from `start`, iterating over the items until you reach the end condition. If needed, flesh it out to a separate iterator struct.

src/tables/feature_index.rs
```diff
@@ -0,0 +1,790 @@
...
+    /// The OSM id of the feature `r` refers to -- `osm_id * 10 + {1,2,3}`
+    /// for {node,way,relation}, same encoding as `Feature.id` elsewhere in
+    /// this codebase (see `make_feature_id`/`feature_to_osm_id` in
+    /// `pipeline::osm::assemble`). O(1), array read only -- no decode.
+    pub fn feature_id(&self, r: LocalFeatureRef) -> u64 {
+        u64::from_le(self.id[r.0 as usize])
+    }
+
+    /// Fully decodes the feature `r` refers to. The only operation on
+    /// this index that costs a protobuf decode -- callers should only do
```
brawer: Uhm, match_mask is already checked by `query()` above, and callers can't score a feature without accessing its contents (modulo the hydration story from the design doc, but that's speculative if we ever build it out). End the sentence after "decode", remove the part about match_mask and scoring.

src/tables/feature_index.rs
```diff
@@ -0,0 +1,790 @@
...
+    /// this for candidates that survive `MatchMask` filtering (and
+    /// typically, scoring).
+    pub fn get_feature(&self, r: LocalFeatureRef) -> Result<Feature> {
+        let i = r.0 as usize;
+        let start = usize::try_from(u64::from_le(self.starts[i]))?;
+        let end = usize::try_from(u64::from_le(self.starts[i + 1]))?;
+        Feature::decode(&self.blob[start..end]).context("failed to decode Feature")
```
brawer: In the error message, indicate which LocalFeatureRef failed to decode. Make sure we don't construct the error message only in case of actual failures, not for every feature being looked up — there's a variant of context() that takes a closure.

src/tables/feature_index.rs
```diff
@@ -0,0 +1,790 @@
...
+/// (features are looked up by [LocalFeatureRef] position, not by id).
+struct StorageWriter {
+    path: PathBuf,
+    tmp_path: PathBuf,
+    writer: BufWriter<File>,
+
+    starts_path: PathBuf,
+    starts_writer: BufWriter<File>,
+
+    blob_path: PathBuf,
```
brawer: blobs_path, blobs_writer — we write many million blobs, not just 1

src/tables/feature_index.rs
```diff
@@ -0,0 +1,790 @@
...
+    blob_writer: BufWriter<File>,
+
+    entries_count: u64,
+}
+
+impl StorageWriter {
```
brawer: I think "FeatureStorageWriter" would better capture what this is doing; every writer writes to storage.

src/tables/feature_index.rs
```diff
@@ -0,0 +1,790 @@
...
+}
+
+/// Writer for the inverted-index file: `coverage_cell_id` and `packed`
+/// arrays as described in the module documentation. Unlike
+/// [crate::tables::BlobTable]/[crate::tables::CoordTable]'s writers,
+/// duplicate keys are expected and allowed (only a *decreasing* key is
+/// rejected) -- a single S2 cell is routinely covered by several
+/// features, and a feature with non-point geometry can itself contribute
+/// more than one entry.
+struct InvertedWriter {
```
brawer: InvertedIndexWriter

src/tables/feature_index.rs
```diff
@@ -0,0 +1,790 @@
...
+    path: PathBuf,
+    tmp_path: PathBuf,
+    writer: BufWriter<File>,
+
+    packed_path: PathBuf,
```
brawer, line 633: I think "features_path", and "features_writer" would better bring across what is being written. Maybe add a short comment, like // serialized Feature protos

src/tables/feature_index.rs
```diff
@@ -0,0 +1,790 @@
...
+    ops::RangeInclusive,
+    path::{Path, PathBuf},
+    time::SystemTime,
+};
+
+/// Size of each file's header, in bytes: see the "File format" section above.
+const HEADER_SIZE: usize = 8 * 8;
+
+const STORAGE_FILE_SIGNATURE: &[u8; 8] = b"featstg0";
+const INVERTED_FILE_SIGNATURE: &[u8; 8] = b"featinv0";
```
brawer: INVERTED_INDEX_FILE_SIGNATURE

src/tables/feature_index.rs
```diff
@@ -0,0 +1,790 @@
...
+    mem::size_of,
+    ops::RangeInclusive,
+    path::{Path, PathBuf},
+    time::SystemTime,
+};
+
+/// Size of each file's header, in bytes: see the "File format" section above.
+const HEADER_SIZE: usize = 8 * 8;
+
+const STORAGE_FILE_SIGNATURE: &[u8; 8] = b"featstg0";
```
brawer: FEATURES_FILE_SIGNATURE

src/tables/feature_index.rs
```diff
@@ -0,0 +1,790 @@
...
+const INVERTED_FILE_SIGNATURE: &[u8; 8] = b"featinv0";
+
+/// An opaque reference to a feature's position in an [OsmFeatureIndex]'s
+/// feature storage. Cheap to copy and hold onto (it's just an array
+/// index); not meaningful outside the [OsmFeatureIndex] that produced it,
+/// and not an OSM id -- see [OsmFeatureIndex::feature_id].
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
```
brawer: Would we gain anything from #[repr(transparent)] on LocalFeatureRef, or would this only matter for interfacing with C? If it makes no difference in a pure Rust pipeline, let's keep it as is. (Don't just make the change, please go find out).

src/tables/feature_index.rs
```diff
@@ -0,0 +1,790 @@
...
+pub struct LocalFeatureRef(u32);
```
brawer, line 114: Somewhere in the writer, we should panic (or return an error, doesn't really matter) if we get more than u32::MAX features. With the current pruning, we have 500M features, so u32 is no problem. But a future change to the pruning code — accidental or intentional, who knows what this pipeline will do in the future — may make this go beyond 2^32, so we should fail (and then change this).

src/tables/feature_index.rs
```diff
@@ -0,0 +1,790 @@
...
+                ))
+                .build()?;
+        let sorted2 = sorter2.sort(pass2_records.map(std::io::Result::Ok))?;
+        let mut inverted_writer = InvertedWriter::create(&inverted_path)?;
+        for item in sorted2 {
+            let (cell_id, match_mask, local_index) = item?;
+            inverted_writer.write(cell_id, match_mask, local_index)?;
+        }
+        inverted_writer.close()?;
+        remove_file(&pass2_spool_path)?;
```
brawer, line 263: Before removing the spool file (which is good), please log the size, so we can see from the logs how much space was used on disk.

src/tables/feature_index.rs
brawer: Investigated: no gain here. `#[repr(transparent)]` guarantees the struct has the exact same ABI/layout as its single field -- that matters for FFI (C interop) and for safely transmuting between the wrapper and inner type via crates like `bytemuck`/`zerocopy`. `LocalFeatureRef` never crosses an FFI boundary and is never transmuted; it's a purely internal opaque handle used only within this module and by callers that just pass it back in. Rust's default (unspecified) layout for a one-field tuple struct already has the same size as the field in practice (rustc has no reason to pad a single-field wrapper), so `repr(transparent)` wouldn't change generated code here -- it would only turn that incidental fact into a documented guarantee, which nothing in this pure-Rust pipeline relies on. Left as is; added a doc comment on `LocalFeatureRef` recording this reasoning so it doesn't need re-litigating later.

src/tables/feature_index.rs
brawer: I think this comment landed on the wrong field -- `InvertedIndexWriter`'s `packed_path`/`packed_writer` (this struct) hold the packed `(match_mask, local_index)` `u64` values, not serialized `Feature` protos, so renaming them to `features_path`/`features_writer` would actually be misleading. I suspect you meant the analogous fields on `FeatureStorageWriter` (`blob_path`/`blob_writer`, a few hundred lines up) -- those *do* hold `Feature.encode_to_vec()` bytes. I renamed those to `blobs_path`/`blobs_writer` per your other comment on that struct (plural, since it's many blobs), and left this struct's fields as `packed_path`/`packed_writer` with a short comment clarifying they're not Feature protos. Let me know if you'd rather I rename these too, and to what.

src/tables/feature_index.rs
brawer: Thanks for the research, and the nice summary!

src/tables/feature_index.rs
```diff
@@ -79,36 +88,69 @@ use std::{
...
+/// `centroid_s2_cell_id`. A named struct (rather than a raw tuple, as
+/// e.g. [crate::tables::BlobTable::create] sorts) so it can derive
+/// [DeepSizeOf]: `fti_bytes`' actual heap size varies a lot between a bare
+/// node and a feature with a large polygon geometry, so the sort buffer
+/// needs to budget by real memory use ([MemoryLimitedBufferBuilder]), not
+/// item count.
+#[derive(Clone, Debug, DeepSizeOf, Deserialize, Eq, Ord, PartialEq, PartialOrd, Serialize)]
+struct Pass1Item {
+    centroid_s2_cell_id: u64,
+    fti_bytes: Vec<u8>,
```
brawer, line 144: Add a comment so it's clear this is a serialized FeatureToIndex proto; from "fti_bytes" alone, this may not obvious to a reader.

src/tables/feature_index.rs
```diff
@@ -79,36 +88,69 @@ use std::{
...
+/// One item [OsmFeatureIndex::create]'s pass 2 external-sorts by
+/// `coverage_cell_id` -- one per entry in a feature's
+/// `coverage_s2_cell_id`. Fixed-size (unlike [Pass1Item]), so the sort
+/// buffer can budget by a plain item count ([LimitedBufferBuilder])
+/// instead of [DeepSizeOf].
+#[derive(Clone, Copy, Debug, Deserialize, Eq, Ord, PartialEq, PartialOrd, Serialize)]
+struct Pass2Item {
+    coverage_cell_id: u64,
+    match_mask: u32,
+    local_index: u32,
```
brawer: Can this be of type LocalFeatureRef instead of u32, for clarity? (Please double-check and push back if that would be wrong).

src/tables/feature_index.rs
```diff
@@ -133,63 +175,83 @@ impl<'a> OsmFeatureIndex<'a> {
...
+                .with_buffer(MemoryLimitedBufferBuilder::new(
+                    EXTERNAL_SORT_CHUNK_BYTES as u64,
                 ))
                 .build()?;
         let sorted1 = sorter1.sort(fti.map(|f| {
-            let key = f.centroid_s2_cell_id;
-            std::io::Result::Ok((key, f.encode_to_vec()))
+            std::io::Result::Ok(Pass1Item {
+                centroid_s2_cell_id: f.centroid_s2_cell_id,
+                fti_bytes: f.encode_to_vec(),
```
brawer, line 204: Add comment to indicate fti_bytes is a serialized FeatureToIndex proto

src/tables/feature_index.rs
brawer: Ah, now I get it. Thanks for havng added the comment, this is helpful.

src/tables/feature_index.rs
brawer: Double-checked -- this is correct, not just cosmetically nice: pass 1's `local_index` *is* exactly the value `LocalFeatureRef` wraps (a feature's position in feature storage), just before it's ever handed to a caller, so typing `Pass2Item::local_index` as `LocalFeatureRef` is the same concept end to end, not a coincidental overlap. It also closes a real gap: `InvertedIndexWriter::write(cell_id, match_mask, local_index)` took two adjacent bare `u32` params with no compile-time protection against swapping them. Made the change (`LocalFeatureRef` now also derives `Ord`/`PartialOrd`/`Serialize`/`Deserialize` so it can be a `Pass2Item` field) in f1471dd.

brawer: Double-checked -- this is correct, not just cosmetically nice: pass 1's local_index *is* exactly the value LocalFeatureRef wraps (a feature's position in feature storage), just before it's ever handed to a caller, so typing Pass2Item::local_index as LocalFeatureRef is the same concept end to end, not a coincidental overlap. It also closes a real gap: InvertedIndexWriter::write(cell_id, match_mask, local_index) took two adjacent bare u32 params with no compile-time protection against swapping them. Made the change (LocalFeatureRef now also derives Ord/PartialOrd/Serialize/Deserialize so it can be a Pass2Item field) in f1471dd.

brawer (commented):
LGTM

PR #660, opened 2026-08-13
Build OsmFeatureIndex in import_osm(), still unconsumed

brawer:
PR 2 of the staged sequence tracked in #655.

New `src/pipeline/osm/index.rs`: merges an `Assembly`'s four
`RecordReader`s (nodes, ways, leaf_relations, super_relations) via one
producer thread per reader — each an independent LZ4 stream, decoded
concurrently — feeding a shared multi-producer channel into
`OsmFeatureIndex::create()`. Deliberately not `rayon::par_bridge()`: the
expensive part (LZ4 decompression) happens inside `RecordReader::iter()`
itself, which is inherently sequential per stream, so bridging a single
reader's iterator wouldn't parallelize anything — four independent
streams decoding concurrently is the actual win.

Wired into `import_osm()` right after `assemble::assemble()`, result
discarded for now (`let _osm_index = ...`), the same way `Assembly`
itself is already discarded today. This is the first exercise of the
index against real PBF data, before `conflate`/`suggest_edits` (PR 4/5)
depend on it.

Also adds `OsmFeatureIndex::exists(path)`, mirroring
`FilteredFeatureStore::exists`, so this step memoizes the same way other
pipeline stages do.

## Testing
- `cargo test --lib` (241 passed), `cargo test --test integration_test`
  fmt --check` clean.
- Manually verified end-to-end against the built CLI binary + test
  fixtures (not just `cargo test`): a full `osm-diffs run` produces valid
  `osm-features.index` + `osm-features.index.inverted` files, and a
  second run correctly skips rebuilding them (unchanged mtime, no
  duplicate "pass 1 → pass 2 spool file" log line).

### DIFF
--- a/src/pipeline/osm/index.rs
+++ b/src/pipeline/osm/index.rs
@@ -30,9 +30,12 @@ use std::thread;
 /// bridging a single reader's iterator would just serialize on the
 /// decompressor anyway, adding rayon dispatch overhead for no gain. Four
 /// independent streams decoding concurrently is the actual parallelism
-/// win here. (If that's ever not enough, the ceiling isn't fixed at
-/// four -- `assemble_nodes`/`assemble_ways`/etc. could shard their own
-/// `RecordWriter` output into several files instead of one each.)
+/// win here. Not optimizing further than that for now -- worth checking
+/// whether this is an actual bottleneck first. If it is, sharding
+/// `assemble_nodes`/`assemble_ways`'s own `RecordWriter` output into more
+/// files would raise the ceiling above four; `leaf_relations`/
+/// `super_relations` likely wouldn't benefit the same way, since
+/// OpenStreetMap has comparatively few relations.
 pub fn build_index<'a>(
     assembly: &Assembly,
     progress: &MultiProgress,
@@ -54,7 +57,7 @@ pub fn build_index<'a>(
         progress,
         "osm.index                   ",
         total,
-        "records → index",
+        "features → index",
     );
 
     let mut create_result: Option<Result<OsmFeatureIndex<'a>>> = None;
--- a/src/pipeline/osm/index.rs
+++ b/src/pipeline/osm/index.rs
@@ -1,0 +1,94 @@
+//! Builds an [OsmFeatureIndex] from an [Assembly]'s four `RecordReader`s.
+//!
+//! Not yet consumed anywhere -- `import_osm` in `mod.rs` calls this and
+//! discards the result, the same way it already discards `Assembly`
+//! itself, until `conflate`/`suggest_edits` are ready to use it. This
+//! exists so the index-building step gets exercised against real PBF
+//! data -- and can be watched with actual production-scale metrics --
+//! before anything downstream depends on it.
+
+use super::assemble::Assembly;
+use crate::make_progress_bar;
+use crate::tables::{FeatureToIndex, OsmFeatureIndex, RecordReader};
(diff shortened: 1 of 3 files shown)


src/pipeline/osm/index.rs
```diff
@@ -0,0 +1,94 @@
...
+/// decode their own stream and send decoded `FeatureToIndex` records down
+/// a shared multi-producer channel; `OsmFeatureIndex::create` drains it.
+///
+/// Deliberately not `rayon::par_bridge()`: the expensive part (LZ4
+/// decompression) happens *inside* each `RecordReader::iter()` call while
+/// pulling the next record, which is inherently sequential per stream --
+/// bridging a single reader's iterator would just serialize on the
+/// decompressor anyway, adding rayon dispatch overhead for no gain. Four
+/// independent streams decoding concurrently is the actual parallelism
+/// win here. (If that's ever not enough, the ceiling isn't fixed at
```
brawer: Remove "ever" from this comment, it sounds hyperbolic. Even my cheap laptop has 10 CPU cores, so our current 4-way parallelism isn't so great. Especially since only two of the input files are large (nodes, ways), whereas the two inputs for leaf and super relations don't really matter because OpenStreetMap has only a few of these. Anyhow, before talking about optimizing this, let's see whether this is an actual performance bottleneck. So, dear Claude, just rephrase your comment to make it sound less hyperbolic about what you built, thank you. :-)

src/pipeline/osm/index.rs
```diff
@@ -0,0 +1,94 @@
...
+        ("ways", &assembly.ways),
+        ("leaf_relations", &assembly.leaf_relations),
+        ("super_relations", &assembly.super_relations),
+    ];
+    let total: u64 = sources.iter().map(|(_, r)| r.len() as u64).sum();
+    let progress_bar = make_progress_bar(
+        progress,
+        "osm.index                   ",
+        total,
+        "records → index",
```
brawer: "features → index"

src/pipeline/osm/index.rs
```diff
@@ -0,0 +1,94 @@
...
+    );
+
+    let mut create_result: Option<Result<OsmFeatureIndex<'a>>> = None;
+    thread::scope(|s| -> Result<()> {
+        let (tx, rx) = sync_channel::<FeatureToIndex>(1024);
+        let progress_bar = &progress_bar;
+
+        let producers: Vec<_> = sources
+            .into_iter()
+            .map(|(name, reader)| {
```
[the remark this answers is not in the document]
brawer: This warning from CodeQL looks like a false positive. Filed a CodeQL bug in https://github.com/github/codeql/issues/22342.

src/pipeline/osm/index.rs
brawer: Fixed in 011aa91.

src/pipeline/osm/index.rs
brawer: Fixed in 011aa91.

src/pipeline/osm/index.rs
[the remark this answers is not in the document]
brawer: Confirmed a false positive -- `name` is used inside the `with_context` closure a couple lines down (`format!("failed to decode a FeatureToIndex record from {name}")`). Thanks for filing the upstream CodeQL bug; no code change needed here.

brawer (commented):
LGTM

PR #661, opened 2026-08-13
Add OsmCandidate + score_osm_candidate, additive

brawer:
PR 3 of the staged sequence tracked in #655.

New `OsmCandidate<'a>` (`matchers/mod.rs`): pairs a `tables::Feature`
with the `StringPool` needed to resolve its tag ids — kept as its own
type (rather than threading `&Feature`/`&StringPool` through `Matcher`
separately) so a future hydration step (e.g. a pre-decoded
`geo::Geometry` for Hausdorff-distance scoring, or a pre-tokenized name
for Token Sort Ratio matching) can attach fields here without another
trait-signature change. `tags()` is cheap; `geometry()` decodes WKB and
is not, so callers should only reach for it when they actually need the
shape.

`trait Matcher` gains a new `score_osm_candidate(&self, &OsmCandidate)`
method, alongside the existing `Place`-based `score`/`suggest_edit` —
not replacing them yet, so `conflate.rs`/`edits.rs` don't need to change
in this PR at all (they still go through `Place`/`PlaceIndex`). It has
to be a trait method rather than an inherent method on `PoiMatcher`:
`conflate`/`edits` call through `Box<dyn Matcher>`, so a later PR that
wires this in can't reach an inherent method that isn't on the trait
object.

`PoiMatcher::score_osm_candidate` mirrors `score`'s `brand:wikidata`
logic, but checks the (cheap) tag match before decoding the (not cheap)
geometry — `score`'s existing unconditional-then-check order made sense
when `Place`'s distance was free to compute; `Feature`'s isn't. Since
`Feature` has no precomputed position (unlike `Place`'s `s2_cell_id`),
distance is computed from the decoded geometry's centroid;
`geo_point_distance`/`geo_point_distance_score` added alongside the
existing `Place`-based `distance`/`distance_score` for this. Also pulled
out `find_brand_wikidata()`, shared by `for_place`/`score`/
`score_osm_candidate` instead of duplicating the same tag scan three
times, and deduplicated a stray inline `EARTH_RADIUS_METERS` constant
into one module-level one.

## Testing
- Ported `test_poi_matcher`'s `CH_CLOTHES`/`CH_KIOSK` fixtures to
  `Feature`/`StringPool` (reusing the same real-world coordinates via
  their `s2_cell_id`, so this exercises the WKB-decode-then-centroid
  path against genuine positions, not synthetic ones), plus direct unit
  tests for `OsmCandidate::tags`/`geometry` and
  `geo_point_distance_score`.
- `cargo test --lib` (246 passed), `cargo test --test integration_test`
  fmt --check` clean.

### DIFF
--- a/src/matchers/mod.rs
+++ b/src/matchers/mod.rs
@@ -1,12 +1,21 @@
 //! Logic for matching AllThePlaces with OpenStreetMap features.
 
 use crate::places::Place;
+use crate::tables::{Feature, StringPool};
+use anyhow::{Context, Result};
 use deepsize::DeepSizeOf;
+use geo_traits::to_geo::ToGeoGeometry;
 use s2::{cellid::CellID, point::Point, s1::ChordAngle};
 use serde::{Deserialize, Serialize};
 use std::sync::LazyLock;
+use wkb::reader::read_wkb;
 
 mod poi_matcher;
+
+/// Mean Earth radius, in meters -- used throughout this module to turn
+/// S2 `Angle`s (dimensionless, on the unit sphere) into real-world
+/// distances.
+const EARTH_RADIUS_METERS: f64 = 6_371_000.0;
 
 /// A bitmask to speed up the matching of AllThePlaces with OpenStreetMap.
 ///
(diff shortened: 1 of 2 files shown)


brawer (commented):
LGTM

PR #664, opened 2026-08-13
Log OsmFeatureIndex::create()'s counts, sizes, and phase timing

brawer:
Follow-up to #660, prompted by manually testing it against the full OSM
planet (see discussion on #655): the only way to learn the feature
count, or how long pass 2 took, was to read them off the terminal's
indicatif progress bar (for the count) or infer them from filesystem
mtimes across two different clocks (for timing) — which is exactly how
a UTC-vs-local-time mixup during that investigation briefly looked like
a 2-hour pass 2, when it was actually ~14 seconds. None of that should
require reading tea leaves; it should be in the log.

- `OsmFeatureIndex::create()` now logs two lines: pass 1's completion
  (extended with `elapsed_seconds` and `feature_count`, alongside the
  existing pass-2-spool-file `bytes`/`path`) and a new `"done"` line
  with `pass1_seconds`/`pass2_seconds`/`total_seconds`, `feature_count`,
  `inverted_index_entries`, and both output files' byte sizes.
- `FeatureStorageWriter`/`InvertedIndexWriter` gain a small
  `entries_count()` getter each, so `create()` can read the count
  already being tracked internally rather than duplicating a counter.
- `build_index()` now logs when it skips rebuilding because
  `OsmFeatureIndex::exists(out)` already holds, matching the
  `"skipping build_coverage, already found ..."` pattern `coverage.rs`
  already uses — previously this path was silent, indistinguishable in
  the log from "there was nothing to skip".
- Corrected a code comment's feature-count estimate from a stale "500M"
  guess to the actual measured 162,812,789 (2026-08-03 planet
  snapshot).

## Testing
- Manually verified against the CLI + test fixtures: both new log
  lines fire with correct values (`feature_count`/sizes matching the
  on-disk files), and the skip-log fires when forcing `import_osm()`
  past its older, outer memoization gate to actually reach this code
  on a second run.
- `cargo test --lib` (241 passed), `cargo test --test integration_test`
  fmt --check` clean.

### DIFF
--- a/src/pipeline/osm/index.rs
+++ b/src/pipeline/osm/index.rs
@@ -43,6 +43,7 @@ pub fn build_index<'a>(
     out: &Path,
 ) -> Result<OsmFeatureIndex<'a>> {
     if OsmFeatureIndex::exists(out) {
+        log::info!("skipping OsmFeatureIndex build, already found {:?}", out);
         return OsmFeatureIndex::open(out);
     }
 
--- a/src/tables/feature_index.rs
+++ b/src/tables/feature_index.rs
@@ -82,7 +82,7 @@ use std::{
     mem::size_of,
     ops::RangeInclusive,
     path::{Path, PathBuf},
-    time::SystemTime,
+    time::{Instant, SystemTime},
 };
 
 /// Size of each file's header, in bytes: see the "File format" section above.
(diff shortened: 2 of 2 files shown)


brawer (commented):
LGTM

PR #665, opened 2026-08-13
Cut conflate over to OsmFeatureIndex

brawer:
PR 4 of the staged sequence tracked in #655. This is the PR the whole
effort is actually about — real OSM geometry in `conflated.parquet`.

`conflate()` no longer touches `Place`/`PlaceIndex` or `FeatureStore` on
the OSM side: it takes a `&tables::OsmFeatures` (new bundle type — an
`OsmFeatureIndex` plus the `StringPool` needed to resolve its tag ids,
built independently by different pipeline stages but almost always used
together) instead of `osm: &Path` + `osm_store: &dyn FeatureStore`.
`ConflatedFeature.osm` is now `Option<Feature>`. `produce_rows` queries
the new index, decodes each surviving candidate via `get_feature()`, and
scores it with `PoiMatcher::score_osm_candidate` (added in #661) instead
of the `Place`-based `score()`. This also deletes the entirely-dead
`osm_store.get_node/way/relation` block conflate carried since #656 —
its results were always discarded; the TODO above it ("use a custom
function instead of try_from, so we can pass a reference to FeatureStore
to extract the geometry") is exactly what this PR does, just via
`OsmFeatureIndex` instead of `FeatureStore`.

**writer.rs**: `TryFrom<ConflatedFeature> for ParquetRow` becomes
`ParquetRow::from_conflated(cf, osm_strings)` — an inherent function can
take the extra `StringPool` parameter a trait impl can't.
`osm_shape_wkb` is now `feature.geometry_wkb`, unchanged — already valid
OGC Simple Features WKB, no more reconstructing a synthetic point from
an S2 cell. The GeoParquet-style schema already had a real
`osm.geometry` column; only the value was ever wrong.

**import_osm()** returns `(PathBuf, OsmFeatures<'a>)` instead of
`(PathBuf, Box<dyn FeatureStore>)` — `osm.parquet`'s path is kept
(`suggest_edits`, PR 5, still needs it), the `FeatureStore` side is
dropped now that nothing needs it. Its own memoized-shortcut path now
also checks `OsmFeatures::exists()` before taking the fast path.

**tables::OsmFeatures { index, strings }**: `exists()`/`open()`/
`modified()` (the latter the max of both components' own `modified()`,
so a caller doing its own staleness check treats the pair as one unit —
conflate uses this right away). `OsmFeatureIndex` itself stays
`StringPool`-agnostic on purpose — it doesn't need one for its own job,
only callers decoding tags do, same reasoning as `OsmCandidate` pairing
`&Feature` with `&StringPool` one level down. `StringPool` gained the
matching `modified()` (it already held onto its `File`, just as
`_file`, unused until now).

**Old-path fallout**, marked `#[allow(unused)]` rather than deleted (out
of scope for this PR — `FeatureStore`/`FilteredFeatureStore` are still
used by the rest of the old path, removal is PR 6):
`FeatureStore::get_node/get_way/get_relation`,
`FilteredFeatureStore::open`, `FilteredFile::feature_index`.

## Testing

Verified against the CLI + test fixtures, not just `cargo test`: all 3
real `brand:wikidata` matches in the fixture (Tchibo/MediaMarkt/Denner,
the same candidates identified while planning #655) now come back as
**POLYGON** in `conflated.parquet`'s `osm.geometry` column, not the old
synthetic POINT — checked via DuckDB's spatial extension, which reads
the file's GeoParquet-style metadata natively:

```
┌─────────┬───────────┬────────────┬────────────────┐
│  type   │    id     │   spider   │ osm_geom_type  │
├─────────┼───────────┼────────────┼────────────────┤
│ way     │ 608979139 │ tchibo     │ POLYGON        │
│ way     │ 737021556 │ mediamarkt │ POLYGON        │
│ way     │ 737021557 │ denner_ch  │ POLYGON        │
└─────────┴───────────┴────────────┴────────────────┘
```

Tags/changeset/version cross-checked against the raw fixture data.
ATP-side geometry unaffected (still POINT, as expected — ATP stays on
`Place`). Second run confirmed `conflated.parquet` is correctly
memoized (unchanged mtime).

`cargo test --lib` (246 passed), `cargo test --test integration_test`
fmt --check` clean.

### DIFF
--- a/src/pipeline/conflate/writer.rs
+++ b/src/pipeline/conflate/writer.rs
@@ -1,4 +1,4 @@
-use anyhow::{Ok, Result};
+use anyhow::{Context, Ok, Result};
 use arrow::array::{
     ArrayRef, RecordBatch, StructArray,
     builder::{
@@ -8,6 +8,7 @@ use arrow::array::{
 use arrow_buffer::builder::NullBufferBuilder;
 use arrow_schema::{DataType, SchemaRef};
 use deepsize::DeepSizeOf;
+use geo::Centroid;
 use parquet::{
     arrow::{ArrowWriter, arrow_writer::ArrowWriterOptions},
     basic::{Compression, ZstdLevel},
@@ -23,6 +24,7 @@ use std::{
 };
 
 use super::ConflatedFeature;
+use crate::{matchers::OsmCandidate, tables::StringPool};
 
 pub struct ParquetWriter {
     path: PathBuf,
@@ -297,16 +299,40 @@ impl ParquetWriter {
     }
 }
 
-impl TryFrom<ConflatedFeature> for ParquetRow {
-    type Error = anyhow::Error;
-
-    fn try_from(cf: ConflatedFeature) -> Result<Self, Self::Error> {
+impl ParquetRow {
+    /// Replaces the old `TryFrom<ConflatedFeature>` -- unlike a `TryFrom`
+    /// impl, this can take the extra `osm_strings` parameter needed to
+    /// resolve an OSM `Feature`'s tag ids into actual strings.
+    pub(super) fn from_conflated(
+        cf: ConflatedFeature,
+        osm_strings: &StringPool,
+    ) -> Result<ParquetRow> {
         let atp = cf.atp;
         let osm = cf.osm;
-        let s2_cell_id = if let Some(ref osm) = osm {
-            osm.s2_cell_id
-        } else if let Some(ref atp) = atp {
+
+        // Internal sort key. Almost always available for free from `atp`
+        // -- `produce_rows` only ever calls this with `atp: Some(..)`,
+        // since it iterates ATP places outward, so the `osm`-only branch
+        // below is unreached today, kept only so this function stays
+        // correct if a caller ever does construct an OSM-only row (e.g.
+        // a future "suggest creating this as a new OSM feature" flow).
+        // Feature carries no precomputed position the way Place does
+        // (see https://github.com/alltheplaces/osm-diffs/issues/662), so
+        // that branch has to decode geometry and compute a centroid --
+        // fine for a today-unreached path, not fine to do unconditionally.
+        let s2_cell_id = if let Some(ref atp) = atp {
             atp.s2_cell_id
+        } else if let Some(ref feature) = osm {
+            let candidate = OsmCandidate {
+                feature,
+                strings: osm_strings,
+            };
+            let centroid = candidate
+                .geometry()?
+                .centroid()
+                .context("OSM feature geometry has no centroid")?;
+            let ll = s2::latlng::LatLng::from_degrees(centroid.y(), centroid.x());
+            s2::cellid::CellID::from(ll).0
         } else {
             anyhow::bail!("ConflatedRow must not have atp and osm both None")
         };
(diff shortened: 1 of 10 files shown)


src/pipeline/conflate/writer.rs
```diff
@@ -297,16 +299,40 @@ impl ParquetWriter {
     }
 }
 
-impl TryFrom<ConflatedFeature> for ParquetRow {
-    type Error = anyhow::Error;
-
-    fn try_from(cf: ConflatedFeature) -> Result<Self, Self::Error> {
+impl ParquetRow {
+    /// Replaces the old `TryFrom<ConflatedFeature>` -- unlike a `TryFrom`
```
brawer: Remove this comment. It helps to undestand the change, but someone who reads this codebase in the future won't really care how exactly this worked earlier.

src/pipeline/conflate/writer.rs
```diff
@@ -297,16 +299,40 @@ impl ParquetWriter {
...
             atp.s2_cell_id
+        } else if let Some(ref feature) = osm {
+            let candidate = OsmCandidate {
+                feature,
+                strings: osm_strings,
+            };
+            let centroid = candidate
+                .geometry()?
+                .centroid()
+                .context("OSM feature geometry has no centroid")?;
```
brawer: Also log the feature ID, using `context_with()` so we don't construct an error message in the successful case.

brawer (commented):
LGTM

brawer:
## Full-planet run on a memory-constrained VM: success

Ran the full pipeline (`import_atp` → `import_osm` → `conflate`) end to end
against the real planet on a Hetzner cpx22 (2 vCPU / 4 GB RAM) —
deliberately undersized so it can't hold the ~20.5 GB `OsmFeatureIndex` in
RAM, to stress-test the central bet of this design: relying on the OS page
cache instead of an explicit LRU decode cache.

**Result: no OOM, no swap, no crash, no manual intervention.**

### Timings (total ≈ 4h59m, 06:15:39 → 11:14:48 UTC)

| Step | Duration |
|---|---|
| `import_atp` | 2m52s |
| `import_osm` (incl. `OsmFeatureIndex::create`: 26.1 min) | 4h48m4s |
| `conflate` (match: 4.8 min + write: 3.4 min) | 8m14s |

### Memory — the thing actually being tested

Peak RSS 3.16–3.39 GiB on a 3.7 GiB box (systemd's own cgroup accounting
agrees: 3.5 GB peak). During `conflate.match`, **94%+ of RSS was
`rss_file_bytes`** — page-cache-backed mmap pages, not heap — which is
exactly the signal this design is betting on.

### Disk

- Final `OsmFeatureIndex`: 17.91 GiB (feature storage) + 2.63 GiB
  (inverted index) = **20.53 GiB**, matching the pre-implementation
  estimate almost exactly.
- `import_osm` peaked at 173.9 GB disk used but settled at 147.7 GB once
  the step finished — **~26 GB was pure external-sort scratch space**,
  freed automatically once the sort's spill files were merged. Worth
  budgeting for on disk-constrained hosts, even though it's transient.
- 163,030,988 OSM features indexed; `conflated.parquet` output:
  1,722,945 rows, 144 MB.

### Geometry-build failures (real OSM data-quality issues)

48 way warnings, 272 relation errors on this planet snapshot
(`replication_timestamp=2026-08-10T00:00:00Z`). Relations fail much more
often proportionally than ways, and log at `ERROR` rather than `WARN`.
Planning a follow-up to #636 with the details.

### Artifacts

Raw `pipeline.log`/`vmstat.log`/`disk.log`/`sysinfo.txt` kept locally
(`scripts/test-branch-on-hetzner/logs/pr665-memtest/`, gitignored) before
tearing the VM down.

PR #668, opened 2026-08-14
Add scripts/test-branch-on-hetzner: automate testing branches on Hetzner Cloud VMs

brawer:
## What

`scripts/test-branch-on-hetzner/cloud_test.py`: spins up a Hetzner Cloud VM, builds a given git branch on it **natively** (no QEMU emulation), runs the pipeline against it detached via `systemd-run`, and pulls back logs — one command instead of repeating the manual VM setup by hand.

```
cloud_test.py up --name pr665-m1 --branch <branch> --ssh-key <key>
cloud_test.py status --name pr665-m1
cloud_test.py logs --name pr665-m1
cloud_test.py destroy --name pr665-m1
```

Subcommands: `up` / `create` / `deploy` / `start` / `status` / `fio` / `sysinfo` / `logs` / `stop` / `destroy` / `list`. Full details in [`scripts/test-branch-on-hetzner/README.md`](scripts/test-branch-on-hetzner/README.md), also reorganized [`scripts/README.md`](scripts/README.md) now that this directory covers more than release engineering.

## Why

Grew out of manually testing #665 (`OsmFeatureIndex`/`conflate`) on three Hetzner VMs, which surfaced the same category of setup friction repeatedly: a replacement VM landing on the wrong OS image (only caught by comparing `uname -a` across machines), a default file-descriptor limit too low for a large external sort, a Hetzner volume-size quota, and a container build that silently targeted the wrong CPU architecture when built locally on Apple Silicon (fine on native arm64, useless on the x86_64 target — followed by an emulated rebuild that also OOM-killed the C++ compiler under QEMU). None of this was specific to what #665 changes — it's generic friction in "get a branch running on a cloud VM and watch it," worth solving once given large rollouts call for repeating this at multiple stages (see #667, which tracks this in full).

## Testing

Shaken down end to end against the real Hetzner API before opening this PR: `create` → `sysinfo` → `fio` → `deploy` → `start` → verified actually running (`systemctl status`, `pipeline.log`/`vmstat.log`/`disk.log` growing) → `stop` → `status` → `logs` → `destroy` → confirmed via `list` that nothing was left running. Found and fixed 4 real bugs along the way:

1. `--server`/`--location` are mutually exclusive on `hcloud volume create`.
2. `hcloud_json()` swallowed command + stderr on failure, contradicting the tool's own "always show what broke" premise.
[...]

PR #670, opened 2026-08-14
Add scripts/test-branch-on-macos: quick local iteration

brawer:
## What

`scripts/test-branch-on-macos/test_macos.py`: build the current checkout, run the pipeline against a workdir, and monitor `vm_stat`/RSS alongside it — one command instead of re-typing the same monitoring loop by hand.

```
test_macos.py run --workdir /tmp/osm-diffs-workdir [--clean] [--skip-build]
test_macos.py build
```

## Why

Much smaller than #668 (`test-branch-on-hetzner/`) on purpose — no VM lifecycle, no branch to clone, plain `cargo build --release` rather than matching production's exact toolchain. The point here is fast local iteration on whatever's currently checked out, not comparable-to-production numbers or real hardware. See #669 for the full motivation.

`monitor.sh` fixes a real bug from an earlier ad hoc version of this script used during the PR 665 experiment: it re-resolves `osm-diffs`'s PID via `pgrep` on every 5-second tick instead of once at the top, so it tolerates being started before the process exists yet (or restarting) — the old version would silently produce a useless log ("Invalid process id") for the rest of a run once the PID went stale.

## Testing

Ran end to end against the small test fixtures (same ones `integration_test.rs` uses) — confirmed exit code 0, `pipeline.log` completing through `conflate`, and critically that `macos_monitor.log` shows a real resolved PID with actual `ps` output, not the old failure mode.

Not part of the Rust build — no `cargo test` impact.

Closes #669.


brawer (commented):
LGTM

PR #678, opened 2026-08-16
Cut suggest_edits over to conflated.parquet (PR 5 of #655)

brawer:
PR 5 of the #655 rollout. Design discussed at length in that issue's thread before implementing.

## What changed

`suggest_edits` no longer searches OSM independently — `conflate()` already decided, per ATP feature, which OSM feature (if any) is the same real-world object. Redoing that search in `suggest_edits` too meant paying for the whole candidate scan a second time, and could in principle disagree with conflate's answer. Instead, `suggest_edits` now does one sequential scan over `conflated.parquet` and asks an `EditSuggester` what to change for each already-matched pair.

- New `src/edit_suggesters/` (`EditSuggester` trait + `PoiEditSuggester`), deliberately separate from `Matcher`: matching needs different signals per category to decide *whether* two features are the same object, but diffing tags once a match is settled doesn't care *how* the match was made, and that logic can be shared across categories matching itself needs to tell apart.
- `Matcher::suggest_edit` is gone — no `OsmCandidate`-based replacement was added, since generating edits isn't a matching concern any more.
- `PoiMatcher` no longer borrows the ATP `Place` (that was only needed for the old `suggest_edit`); drops its lifetime parameter as a result.
- `suggest_edits`' `conflated.parquet` reader is a small, purpose-built sequential Arrow/Parquet scan — no caching, no spatial index, since every row is visited exactly once, in file order.
- Deliberately did **not** add a persisted match-score column to `conflated.parquet` — that file is meant to stay a clean, externally-legible data product, and both old and new code use the same "any nonzero score wins" bar today, so this isn't a behavior change.

## Deferred, not in scope here

`conflate()`'s ATP-side read still goes through `PlaceIndex`, so its now-dead query/LRU-cache machinery (nothing else calls `query()`/`cache_stats()` any more) is `#[allow(dead_code)]` rather than deleted — cleanly removing it also means moving that ATP read off `PlaceIndex` onto a plain sequential reader, which is a separate, focused follow-up PR rather than part of this change.

## Testing

`test_pipeline` now asserts on `shops.jsonl`'s actual contents (concrete expected tags/values from the fixture data), not just that the run succeeds — verified it actually catches regressions by temporarily breaking the trustworthy-tag allowlist and confirming the test fails.

### DIFF
--- a/src/edit_suggesters/mod.rs
+++ b/src/edit_suggesters/mod.rs
@@ -1,0 +1,45 @@
+//! Turns a conflated ATP+OSM feature pair -- one that
+//! `crate::pipeline::conflate` has already determined refer to the same
+//! real-world object -- into a suggested OpenStreetMap edit.
+//!
+//! Deliberately separate from `crate::matchers::Matcher`, even though
+//! both are dispatched by category: matching needs different signals
+//! per category to decide *whether* two features are the same object
+//! (e.g. brand for shops; a restaurant or museum matcher would likely
+//! need different signals again). But once a match is settled, deciding
+//! *what to change* -- e.g. diffing `opening_hours` -- doesn't care how
+//! the match was made, and that logic can be shared across categories
+//! that matching itself needs to tell apart. So `EditSuggester`
+//! implementations don't need to mirror `Matcher` implementations
+//! one-for-one; there may end up being far fewer of them.
+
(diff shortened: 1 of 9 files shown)


brawer (commented):
LGTM

PR #679, opened 2026-08-16
tests: assert on conflated.parquet's actual content

brawer:
Extends `test_pipeline` to inspect `conflated.parquet` directly (Arrow/Parquet reading, not a golden-file diff), pinning it against the fixture data's known matches — three shops (Tchibo, MediaMarkt, Denner) that `conflate()` matches by `brand:wikidata`, each asserted to carry real polygon geometry.

Verified via DuckDB against a fresh fixture run before writing the assertions, and confirmed the test actually fails when an expected value is wrong (temporarily broke one, watched it fail, reverted).

### DIFF
--- a/tests/integration_test.rs
+++ b/tests/integration_test.rs
@@ -62,10 +62,7 @@ struct ConflatedOsmSide {
 }
 
 /// Checks `conflate()`'s output against the fixture data's known shops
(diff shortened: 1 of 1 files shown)


brawer (commented):
LGTM

PR #680, opened 2026-08-16
Delete PlaceIndex; conflate() reads ATP via a plain sequential reader

brawer:
Follow-up to #678 (as agreed there): `PlaceIndex`'s query/LRU-cache machinery existed to serve spatial OSM-candidate lookups — a need that went away twice over, once when `conflate()` moved to `OsmFeatureIndex` (#665), again when `suggest_edits` stopped searching OSM at all (#678). Only `open()`/`scan_row_groups()` (`conflate()`'s ATP-side read) were still live, and that read never needed a cache either: every row is visited exactly once, in file order.

## What changed

- New `places::PlaceReader`: `open()` + `total_rows()` + `read_all()`, no `Arc`, no LRU, no query API — replaces `PlaceIndex` entirely (`src/places/place_index.rs` deleted, 538 lines).
- `conflate()`'s ATP read now borrows a plain `&PlaceReader` across its scoped producer thread instead of cloning an `Arc<PlaceIndex>`.
- Also deletes `Place::to_geojson`, dead since #678 moved `suggest_edits`' GeoJSON output to `SuggestedEdit::to_geojson` — a straggler that `PlaceIndex`'s own `#[allow(dead_code)]` didn't cover, since it's a different type in a different file. Clippy flagged it once `PlaceIndex` (and the dead code it was suppressing warnings for) was gone.

Net: -634 / +~200 lines.

## Testing

Verified against the fixture data: `shops.jsonl` is byte-identical before and after this change. Full `fmt`/`clippy --all-targets -D warnings`/`test` all green.


src/places/reader.rs
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    pub fn total_rows(&self) -> usize {
+        self.total_rows
+    }
+
+    /// Reads every row, batch by batch (Arrow's natural row-group-ish
+    /// chunking). The caller processes each batch in parallel with
+    /// Rayon; batches preserve spatial locality (the file is S2-sorted)
+    /// even though nothing here relies on that ordering itself.
```
brawer: As written now, this is a bit misleading. The spatial ordering doesn't have to be exact (that's why we process a single batch in parallel), but a _rough_ spatial ordering is super important for performance — the effective of our memory-mapping story depends on it. Without spatial ordering, we'd be wildly thrashing when processing the full planet on a small machine, where it doesn't fit all into RAM. Might be good to explain this here.

src/places/reader.rs
```diff
@@ -0,0 +1,198 @@
+//! Sequentially reads a `Place`-typed Parquet file (`alltheplaces.parquet`
+//! today -- the only remaining caller is `conflate()`, reading ATP).
+//!
+//! No caching, no spatial index: unlike the `PlaceIndex` this replaces,
```
brawer: No need to talk about history / past approaches, just explain the current code structure.

src/places/reader.rs
```diff
@@ -0,0 +1,198 @@
...
+                .map(|row| extract_place(&batch, row))
+                .collect()
+        }))
+    }
+}
+
+fn extract_place(batch: &RecordBatch, row: usize) -> Result<Place> {
+    Ok(Place {
+        s2_cell_id: get_u64_required(batch, "s2_cell_id", row)?,
+        osm_id: get_u64_optional(batch, "osm_id", row)?.and_then(NonZeroU64::new),
```
brawer, line 70: Are osm_id, osm_changeset and osm_version ever accessed? That would be surprising, given that we're now taking OSM data from OsmFeatureIndex, and PlaceReader is only used for ATP data now. If possible, let's delete them from struct Place. Or is there a reason to keep these fields?

src/places/reader.rs
brawer: Fixed — you're right, that was backwards. Rewrote it to explain that exact order within a batch doesn't matter (that's why Rayon can parallelize it), but the rough spatial ordering across batches does: it keeps conflate()'s per-ATP-feature OsmFeatureIndex queries clustered in nearby regions of the mmap, which is what the page-cache-locality design depends on.

src/places/reader.rs
brawer: Fixed — dropped the PlaceIndex-history framing, just describes the current no-cache-needed structure now.

src/places/reader.rs
brawer: Checked — they're not dead, just not needed for ATP: src/pipeline/osm/old_assemble.rs (the old Place-based OSM path, still alive until #655's PR6) still sets all three when writing OSM-side Place rows to the old osm.parquet. Deleting the fields now would break that compilation.

They'd become genuinely removable once PR6 deletes old_assemble.rs/filter.rs/cover.rs and Place becomes ATP-only for real — flagging it there rather than doing it here, since this PR is scoped to the PlaceIndex→PlaceReader swap.

PR #681, opened 2026-08-16
Delete the old Place-based OSM pipeline (PR 6 of #655)

brawer:
Delete the old Place-based OSM pipeline (PR 6 of #655)

Last stage of the #655 rollout: everything the new OsmFeatureIndex/
conflate/suggest_edits path made redundant.

- Deleted src/pipeline/osm/{filter,old_assemble,cover,coords}.rs,
  src/coverage.rs, and FeatureStore/MockFeatureStore/Node/Way/Relation/
  RelationMember/MemberRole/build_relation_parents/ParentChainIter in
  osm/mod.rs.
- import_osm() no longer takes a coverage: &Path and no longer builds
  osm.parquet -- it now just returns OsmFeatures, the new path's own
  output. conflate()'s staleness check drops coverage too.
- Matcher::score/distance/distance_score deleted -- score_osm_candidate
  is the only scoring path left, now that OSM data never goes through
  Place. Place itself drops osm_id/osm_changeset/osm_version (only
  old_assemble.rs ever set them); places::ParquetWriter's now-always-
  false `osm: bool` parameter goes with it.
- build_coverage's ~40-minute geographic-footprint computation is gone
  entirely: nothing read Coverage's actual content any more (conflate()
  only ever used its mtime for staleness-checking), and pr665's
  memory-pressure run already showed the new path handles the whole
  planet fine without geographic scoping.
- New src/atp/wikidata_ids.rs carries forward the one part of
  coverage.rs still wanted: the set of Wikidata IDs AllThePlaces
  references (brand:wikidata etc.), written to
  workdir/alltheplaces.wikidata-ids as a U64Set. Not consumed by
  anything yet -- kept for a planned future feature (flagging
  OSM-only conflated.parquet rows whose brand ATP tracks elsewhere,
  a sharper "might be a closed store still in OSM" signal than
  emitting every unmatched OSM feature would be). Much cheaper than
  the old build_coverage: a plain tag scan over already-imported ATP
  data, no geographic computation.

Net -3046 lines. Verified against the fixture data: shops.jsonl is
byte-identical before and after; osm.parquet/coverage no longer appear
in the workdir; collect_wikidata_ids runs and logs correctly within
the real pipeline, not just its own isolated test.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


src/matchers/mod.rs
```diff
@@ -283,27 +283,14 @@ impl<'a> OsmCandidate<'a> {
 }
 
 /// Trait for objects that can score a match candidate.
+///
+/// Only scores against [OsmCandidate] (the geometry-aware
```
brawer: Remove this paragraph. No need to explain the history, just the current code.

src/places/mod.rs
```diff
@@ -2,20 +2,21 @@ use crate::matchers::MatchMask;
...
 use serde::{Deserialize, Serialize};
-use std::num::{NonZeroU32, NonZeroU64};
 
 mod reader;
 mod writer;
 
 pub use reader::PlaceReader;
 pub use writer::ParquetWriter;
 
+/// An AllThePlaces feature. (Used to also represent OpenStreetMap
```
brawer: No need to explain the history, just the current code.

src/places/mod.rs
```diff
@@ -2,20 +2,21 @@ use crate::matchers::MatchMask;
...
+/// at one point -- back when OSM went through this same type; OSM now
+/// has its own `tables::Feature`/`OsmFeatureIndex`, so `Place` is
+/// ATP-only. See alltheplaces/osm-diffs#655.)
 #[derive(Debug, DeepSizeOf, Eq, Ord, PartialEq, PartialOrd, Serialize, Deserialize)]
 pub struct Place {
     pub s2_cell_id: u64,
-    pub osm_id: Option<NonZeroU64>,
-    pub osm_changeset: Option<NonZeroU64>,
-    pub osm_version: Option<NonZeroU32>,
     pub source: String,
```
brawer: Rename to "source" to "spider", matching ATP terminology.

src/matchers/mod.rs
brawer: Fixed — trimmed to just describe the current trait.

src/places/mod.rs
brawer: Fixed.

src/places/mod.rs
brawer: Renamed Place.source (and the Parquet column) to spider throughout -- Place::new/deep_clone, places::{reader,writer}, atp::make_place, conflate/writer.rs, poi_matcher.rs's test fixtures. Also had to regenerate tests/test_data/alltheplaces.parquet (a static checked-in fixture with the old column name) -- verified identical content otherwise via DuckDB before/after.

PR #683, opened 2026-08-17
Upload conflated.parquet and pipeline.log to S3; add docs/LOGGING.md

brawer:
Upload conflated.parquet and pipeline.log to S3; add docs/LOGGING.md

Two new uploads alongside the existing tiles upload:

- upload_conflated: same pattern as upload_tiles -- fixed key
  ("conflated.parquet"), overwritten each run.
- upload_logs: keyed by run timestamp ("logs/<run-id>.log"), so weekly
  runs accumulate instead of overwriting -- the whole point being able
  to compare stats (memory, disk, step timings) across runs over time.
  <run-id> reuses pipeline_start_time, the same timestamp already
  embedded into conflated.parquet's own provenance BOM, so a run's log
  and its data output can always be tied back together.
- Uploads happen regardless of whether the pipeline run succeeded:
  run_pipeline now runs the actual step sequence via a new
  run_pipeline_steps() helper, captures its Result, uploads the log
  either way (logging the failure first if there was one, so it's the
  last, easy-to-find line in the archived log), then returns the
  original result. A failed run's log is exactly the one you want
  archived for debugging, not just a successful one's.

Also, per review: upload.rs previously always used a multi-part
upload, appropriate for tiles/conflated.parquet but wasteful for a
small log file. upload_file now picks a single PUT or multi-part based
on the file's size (MULTIPART_THRESHOLD).

Added real test coverage for upload.rs (previously untested): both the
existing tiles path and the two new ones, using mockito against the
actual S3 multipart/PUT request shapes. Refactored S3 configuration
into its own S3Config type, read from the environment as before for
production callers, but injectable directly in tests -- avoids relying
on process-global environment variables, which cargo test's parallel
execution makes an actively racy thing to share between tests (found
this the hard way: the first version of these tests, using an env-var
guard, flaked under parallel execution).

New docs/LOGGING.md explains the JSON Lines log format and where
weekly-run logs land in S3, for someone new to the project. Links two
real examples of what this archive already enables: the #665 PR
comment analyzing a full-planet memory-pressure run, and the #636
data-quality survey -- both written by Claude Code directly from the
logs, no bespoke tooling needed.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/docs/LOGGING.md
+++ b/docs/LOGGING.md
@@ -3,22 +3,22 @@
 `osm-diffs run` writes `pipeline.log` to its `--workdir`, one JSON
 object per line ([JSON Lines](https://jsonlines.org/)) —
 machine-readable rather than free-form text, so a log can be grepped,
-`jq`'d, or loaded into whatever analysis tool you like without first
-having to parse a human-oriented formatter's line-wrapping and colors.
+`jq`’d, or loaded into whatever analysis tool you like without first
+having to parse a human-oriented formatter’s line-wrapping and colors.
 Every record has:
 
 - `timestamp` — RFC 3339, UTC.
-- `level` — `INFO`, `WARN`, `ERROR`, ...
+- `level` — `INFO`, `WARN`, `ERROR`, …
 - `target` — which module logged it (e.g. `osm_diffs::pipeline::conflate`).
 - `message` — the human-readable text.
 - `fields` — present only on records that attach structured data
-  (numbers, an id, ...) instead of just interpolating it into the
-  message text, e.g. a pipeline step's elapsed time and memory
-  snapshot. Omitted entirely on records that don't have any, so plain
+  (numbers, an id, …) instead of just interpolating it into the
+  message text, e.g. a pipeline step’s elapsed time and memory
+  snapshot. Omitted entirely on records that don’t have any, so plain
   log lines stay plain.
 
 See [`src/logging.rs`](../src/logging.rs) for the implementation, and
-[`src/pipeline/mod.rs`](../src/pipeline/mod.rs)'s `log_snapshot` for an
+[`src/pipeline/mod.rs`](../src/pipeline/mod.rs)’s `log_snapshot` for an
 example of a structured record (step name, phase, elapsed time, RSS/
 cgroup memory snapshot — logged at the start and end of every pipeline
 step).
@@ -26,24 +26,22 @@ step).
 ## Where weekly-run logs end up
 
 Every pipeline run uploads its `pipeline.log` to S3 storage at
-`logs/<run-id>.log`, where `<run-id>` is the run's start timestamp —
-the same one embedded into `conflated.parquet`'s own CycloneDX
-provenance BOM (not the release image's SBOM described in
-[`SUPPLY_CHAIN_SECURITY.md`](SUPPLY_CHAIN_SECURITY.md) — a different
-document about a different artifact; see
-[`src/provenance.rs`](../src/provenance.rs) for this one), so a given
-run's log and its data output can always be tied back together. This
-happens regardless of whether the run succeeded — a failed run's log is
+`logs/<run-id>.log`, where `<run-id>` is the run’s start timestamp.
+That timestamp is effectively each production run’s own ID — for
+traceability, it’s also stamped into `conflated.parquet`’s own metadata
+(see [`CONFLATED_OUTPUT.md`](CONFLATED_OUTPUT.md)), so a run’s log and
+its data output can always be tied back together. This happens
+regardless of whether the run succeeded — a failed run’s log is
 exactly the one you want archived for debugging, not just a
-successful one's.
+successful one’s.
 
 ## Why bother
 
-Beyond debugging a single run, having every week's log archived means
+Beyond debugging a single run, having every week’s log archived means
 you can compare stats *across* runs over time — memory/disk usage as
 the planet grows, how a code change shifted step timings, that kind of
 thing. Two real examples of what this data enables, both written by
-[Claude Code](https://claude.com/claude-code) straight from a run's
+[Claude Code](https://claude.com/claude-code) straight from a run’s
 logs, with no other tooling built for the purpose:
 
 - [alltheplaces/osm-diffs#665, comment](https://github.com/alltheplaces/osm-diffs/pull/665#issuecomment-5303068423) —
@@ -53,7 +51,7 @@ logs, with no other tooling built for the purpose:
   explicit decode cache).
 - [alltheplaces/osm-diffs#636](https://github.com/alltheplaces/osm-diffs/issues/636) —
   a survey of real OpenStreetMap data-quality issues, found entirely
-  by combing through `pipeline.log`'s `could not build geometry`
+  by combing through `pipeline.log`’s `could not build geometry`
   warnings and cross-referencing the flagged features against the
   live OSM API.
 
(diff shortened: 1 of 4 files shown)


src/pipeline/upload.rs
```diff
@@ -1,7 +1,27 @@
+//! Uploads pipeline outputs to S3-compatible storage.
+//!
+//! Every upload goes through [upload_file], which decides between a
+//! single PUT and a multi-part upload based on the file's size (see
+//! [MULTIPART_THRESHOLD]) -- multi-part's own per-part minimum makes it
+//! pure overhead for something as small as a log file, but it's still
+//! the right choice for larger files like tiles or `conflated.parquet`.
+//!
+//! All uploads are skipped (not an error) if `S3_ENDPOINT` isn't set --
+//! the established way to disable uploads for a local/dev run.
```
brawer, line 10: Add a comment to document the various environment variables, and what they mean. (Maybe, "pipeline run --help" could also print this what environment variables the tool knows about, but not worth a lot of complexity in the code — if it's trivial, add it to the help message, otherwise just a comment here).

docs/LOGGING.md
```diff
@@ -0,0 +1,61 @@
...
+See [`src/logging.rs`](../src/logging.rs) for the implementation, and
+[`src/pipeline/mod.rs`](../src/pipeline/mod.rs)'s `log_snapshot` for an
+example of a structured record (step name, phase, elapsed time, RSS/
+cgroup memory snapshot — logged at the start and end of every pipeline
+step).
+
+## Where weekly-run logs end up
+
+Every pipeline run uploads its `pipeline.log` to S3 storage at
+`logs/<run-id>.log`, where `<run-id>` is the run's start timestamp —
```
brawer: As it is written now, this paragraph is a bit difficult to read. Maybe end the sentence after "timestamp". In a separate sentence, mention that this timestamp is essentially the ID for our pipeline runs in production, and that we also stamp it (for traceability) into the metadata of conflated.parquet. No need to talk about CycloneDX here, just link to the (not yet existing, see #645 ) docs/CONFLATED_OUTPUT.md.

docs/LOGGING.md
```diff
@@ -0,0 +1,61 @@
+# Logging
```
brawer, line 1: Please use typographic apostrophes, quotes and dashes.

src/pipeline/upload.rs
brawer: Added a doc comment on S3Config listing all five variables and what each means, plus why they're env vars rather than CLI flags (ambient deployment config, not per-invocation). Left --help alone -- wiring 5 non-CLI env vars into clap's help output for documentation purposes only felt like more complexity than it's worth here.

docs/LOGGING.md
brawer: Rewrote it -- shorter sentences, explains the timestamp is effectively each production run's own ID, drops the CycloneDX detail, and links to docs/CONFLATED_OUTPUT.md (not yet written, #645) instead of SUPPLY_CHAIN_SECURITY.md.

docs/LOGGING.md
brawer: Fixed -- converted the whole file to typographic apostrophes/quotes/dashes.

docs/LOGGING.md
```diff
@@ -53,7 +51,7 @@ logs, with no other tooling built for the purpose:
   explicit decode cache).
 - [alltheplaces/osm-diffs#636](https://github.com/alltheplaces/osm-diffs/issues/636) —
   a survey of real OpenStreetMap data-quality issues, found entirely
```
brawer: Remove "real" and "entirely", the current wording sounds a bit too much like a PR text.

docs/LOGGING.md
```diff
@@ -3,47 +3,45 @@
...
-successful one's.
+successful one’s.
 
 ## Why bother
 
-Beyond debugging a single run, having every week's log archived means
+Beyond debugging a single run, having every week’s log archived means
 you can compare stats *across* runs over time — memory/disk usage as
 the planet grows, how a code change shifted step timings, that kind of
 thing. Two real examples of what this data enables, both written by
```
brawer: "Two real examples" --> "Two concrete examples"

brawer: Fixed.

docs/LOGGING.md
brawer: Fixed -- dropped both.

src/pipeline/upload.rs
```diff
@@ -80,68 +101,426 @@ impl<'a> Drop for Upload<'a> {
...
+    fn logs_upload_key_is_run_id_dot_log() -> Result<()> {
+        let mut server = Server::new();
+        let config = test_config(&server);
+
+        let format = time::format_description::well_known::Rfc3339;
+        let t = UtcDateTime::parse("2026-03-04T15:16:17Z", &format)?;
+
+        let mock = server
+            .mock("PUT", "/test-bucket/logs/2026-03-04-15-16-17.log")
+            .match_header("content-type", "application/x-ndjson")
```
brawer: The registered MIME type is "application/jsonl"

src/pipeline/upload.rs
```diff
@@ -80,68 +101,426 @@ impl<'a> Drop for Upload<'a> {
...
+    workdir: &Path,
+    pipeline_start_time: UtcDateTime,
+    progress: &MultiProgress,
+) -> Result<()> {
+    let log_path = workdir.join("pipeline.log");
+    let destination = format!("logs/{}.log", run_id(pipeline_start_time)?);
+    upload_file(
+        &log_path,
+        &destination,
+        "application/x-ndjson",
```
brawer, line 323: Add a TODO comment, something like: We should use an official, IANA-assigned content type for JSON Lines here, but as of August 2026, no consensus has yet been reached on what string to use, so the registration appears to be stalled. See https://github.com/wardi/jsonlines/issues/19. — Also file a GitHub issue to our own repo, so we get remembered to check back in August 2027.

src/pipeline/upload.rs
```diff
@@ -80,68 +101,426 @@ impl<'a> Drop for Upload<'a> {
...
+}
+
+/// Does the actual work for [upload_file], taking an already-resolved
+/// `config` rather than reading the environment itself -- split out
+/// so tests can inject a config pointing at a mock server directly,
+/// instead of going through process-global environment variables
+/// (which `cargo test`'s parallel execution makes an awkward, racy
+/// thing to share between tests).
+fn upload_file_with_config(
+    config: Option<&S3Config>,
```
brawer, line 204: Should we log endpoint/bucket/destination, content type, and file size? If there's some URL convention (or even standard, eg RFC) for objects in S3-compatible storage, it might be nice to just log a single value with that URL; otherwise, separate endpoint/bucket/destination. Also, check if the response after a successful upload includes the final Etag for the object; if we have this value, it might be worth logging it in the same log message.

src/pipeline/upload.rs
brawer: Fixed -- filed alltheplaces/osm-diffs#684 to track checking back once JSON Lines' content-type registration un-stalls, and added a TODO in `upload_logs` referencing it (`c5dc338`).

src/pipeline/upload.rs
brawer: Fixed -- `Upload::finish()` now actually returns the etag (its doc comment already said it did; the code was discarding it), captured from both the single-PUT and multipart paths, and logged in a structured `upload_file: done` record along with an `s3://bucket/key` URL, content type, and byte count. Added an inline comment explaining why `s3://` rather than separate fields or an `https://` URL (`c5dc338`).

brawer (commented):
LGTM

PR #685, opened 2026-08-17
Consolidate external-sort chunk sizes onto EXTERNAL_SORT_CHUNK_BYTES (#657)

### LINKED ISSUE #657: Consolidate external-sort chunk-size constants into EXTERNAL_SORT_CHUNK_BYTES
Several call sites across the pipeline configure \`ext_sort::ExternalSorter\` buffer sizes with their own ad hoc magic numbers, grown over time with no consistent rationale:

- \`BlobTable::create\`: \`LimitedBufferBuilder::new(128 * 1024, true)\`
- \`CoordTable::create\`: \`LimitedBufferBuilder::new(16 * 1024 * 1024, true)\`
- \`assemble_strings\` (pipeline/osm/assemble.rs): \`LimitedBufferBuilder::new(4 * 1024 * 1024, true)\`
- \`conflate::write_conflated\`: \`MemoryLimitedBufferBuilder::new(150_000_000)\`
- \`edits::write_edits\`: \`MemoryLimitedBufferBuilder::new(150_000_000)\`
- \`OsmFeatureIndex::create\` (src/tables/feature_index.rs, added in #656): now uses the new \`pipeline::EXTERNAL_SORT_CHUNK_BYTES\` (512 MB) for both its passes, per review on #656.

There's no good reason for these to be so uneven in size. Once \`EXTERNAL_SORT_CHUNK_BYTES\` exists (landing in #656), migrate the other call sites to use it too, so external-sort chunk sizing is consistent (and tunable in one place) across the pipeline.

Flagged as clean-up, not urgent -- do this on a rainy day, not bundled into unrelated changes.

brawer:
Closes #657.

## Why now

Motivated by a concrete problem: the pr665 Hetzner shakedown hit the
OS's open-file-descriptor limit during external sorting. `ext_sort`
keeps every spilled chunk's file open at once during its final merge, so
chunk count directly determines concurrently-open fds -- and chunk sizes
across this pipeline had grown ad hoc and were wildly uneven (128 KiB up
to 150 MB), several of them small enough to produce a lot of chunks at
planet scale. Consolidating onto one named, tunable constant
(`EXTERNAL_SORT_CHUNK_BYTES`) means there's now a single known knob to
raise if that happens again, instead of hunting down magic numbers
scattered across the codebase.

## What changed

- `BlobTable`, `GeometryTable`, `StringCounts` now take `chunk_bytes`
  explicitly and measure it in actual bytes (`MemoryLimitedBufferBuilder`,
  backed by `DeepSizeOf`) rather than an item count, since their items
  (blobs, strings) vary in size.
- `CoordTable`, `GraphTable`, `U64Set` also take `chunk_bytes`, converted
  to an item count for `LimitedBufferBuilder` since their items are
  fixed-size -- the same pattern `feature_index.rs` already established
  for its own two passes back in #656.
- Single-caller, never-concurrent sites (`StringPool`,
  `conflate::write_conflated`, `atp::import_atp`, `feature_index.rs`'s
  two sequential passes) just use `EXTERNAL_SORT_CHUNK_BYTES` directly --
  no parameter needed since there's nothing to divide it by.

## The part that isn't just a rename

While tracing every call site, I found several pipeline stages
(`prune_relations_pass_1`/`_2`, `prune_ways`, `prune_nodes`, one spot in
`assemble.rs`, and `suggest_edits`'s three output layers) run 2-3
external sorts concurrently on sibling threads within the same
`thread::scope`. Naively giving every call site the full constant would
have multiplied peak memory per stage instead of just cutting file
descriptors. So `EXTERNAL_SORT_CHUNK_BYTES` is treated as a
**per-thread-scope budget**: a lone sorter in its scope gets the full
value; concurrent ones divide it by however many run alongside them (a
local `CONCURRENT_SORTS` constant at each such call site, commented
inline with which sibling threads it's sharing with). This keeps
worst-case peak memory per pipeline stage bounded to roughly one
`EXTERNAL_SORT_CHUNK_BYTES`, regardless of how many sorters happen to run
inside it.

Because the affected table types (`BlobTable`, `CoordTable`, `GraphTable`,
`U64Set`, `StringCounts`) are reusable and called from several different
places with different concurrency, the caller -- who actually knows how
many sibling sorts are running alongside it -- decides how to split the
budget, rather than the table constructor guessing.

## Testing

`cargo build --locked --all-targets`, `cargo test --locked` (236 unit +
4 integration tests, including the full `test_pipeline` end-to-end run),
`cargo fmt --check`, `cargo clippy --locked --all-targets -- -D warnings`
all clean.

### DIFF
--- a/src/pipeline/osm/assemble.rs
+++ b/src/pipeline/osm/assemble.rs
@@ -94,8 +94,8 @@ fn assemble_strings<'a>(
         strings.len() as u64,
         "strings",
     );
-    // Not shared with any sibling thread (unlike several of the sorts in
-    // `prune.rs`), so it gets the full chunk-size budget; see
+    // Not concurrent with any other external sort (unlike several of the
+    // sorts in `prune.rs`), so it gets the full chunk-size budget; see
     // `crate::pipeline::EXTERNAL_SORT_CHUNK_BYTES`. Strings vary in
     // length, so this is measured in actual bytes
     // (`MemoryLimitedBufferBuilder`), not an item count.
--- a/src/pipeline/osm/assemble.rs
+++ b/src/pipeline/osm/assemble.rs
@@ -343,7 +343,7 @@ fn assemble_ways<'a>(
                 geometry_rx.into_iter(),
                 workdir,
                 &ways_in_relations_path,
-                EXTERNAL_SORT_CHUNK_BYTES as u64,
+                EXTERNAL_SORT_CHUNK_BYTES,
             )
         });
 
@@ -520,7 +520,7 @@ fn assemble_leaf_relations<'a>(
                 super_rel_rx.into_iter(),
                 workdir,
                 &super_relations_path,
-                (EXTERNAL_SORT_CHUNK_BYTES / CONCURRENT_SORTS) as u64,
+                EXTERNAL_SORT_CHUNK_BYTES / CONCURRENT_SORTS,
             )?);
             Ok(())
         });
@@ -530,7 +530,7 @@ fn assemble_leaf_relations<'a>(
                 geometry_rx.into_iter(),
                 workdir,
                 &leaf_relations_geometry_path,
-                (EXTERNAL_SORT_CHUNK_BYTES / CONCURRENT_SORTS) as u64,
+                EXTERNAL_SORT_CHUNK_BYTES / CONCURRENT_SORTS,
             )?);
             Ok(())
         });
--- a/src/pipeline/osm/assemble.rs
+++ b/src/pipeline/osm/assemble.rs
@@ -514,13 +514,13 @@ fn assemble_leaf_relations<'a>(
         // is a plain RecordWriter, not a sort) -- split the chunk-size
         // budget between the two so their combined peak memory stays
         // within one EXTERNAL_SORT_CHUNK_BYTES, not double it.
-        const CONCURRENT_SORTS: u64 = 2;
+        const CONCURRENT_SORTS: usize = 2;
         let super_rel_writer = s.spawn(|| {
             super_relations = Some(BlobTable::create(
                 super_rel_rx.into_iter(),
                 workdir,
                 &super_relations_path,
-                EXTERNAL_SORT_CHUNK_BYTES as u64 / CONCURRENT_SORTS,
+                (EXTERNAL_SORT_CHUNK_BYTES / CONCURRENT_SORTS) as u64,
             )?);
             Ok(())
         });
@@ -530,7 +530,7 @@ fn assemble_leaf_relations<'a>(
                 geometry_rx.into_iter(),
                 workdir,
                 &leaf_relations_geometry_path,
-                EXTERNAL_SORT_CHUNK_BYTES as u64 / CONCURRENT_SORTS,
+                (EXTERNAL_SORT_CHUNK_BYTES / CONCURRENT_SORTS) as u64,
             )?);
             Ok(())
         });
--- a/src/pipeline/osm/assemble.rs
+++ b/src/pipeline/osm/assemble.rs
@@ -3,6 +3,7 @@ use crate::{
     geometry::{GeometryBuilder, PolygonFill, build_line, build_ring},
     make_progress_bar,
     matchers::MatchMask,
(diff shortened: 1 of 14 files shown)


src/pipeline/osm/assemble.rs
```diff
@@ -497,11 +509,18 @@ fn assemble_leaf_relations<'a>(
             Ok(())
         });
 
+        // super_rel_writer and leaf_geometry_writer each run their own
+        // external sort concurrently in this scope (leaf_rel_writer above
+        // is a plain RecordWriter, not a sort) -- split the chunk-size
+        // budget between the two so their combined peak memory stays
+        // within one EXTERNAL_SORT_CHUNK_BYTES, not double it.
+        const CONCURRENT_SORTS: u64 = 2;
```
brawer, line 517: Make CONCURRENT_SORTS an usize, for readability. Likewise in other the places being changed.

brawer: Fixed -- CONCURRENT_SORTS is now usize everywhere, dividing EXTERNAL_SORT_CHUNK_BYTES (also usize) directly and casting to u64 once at the end for the chunk_bytes: u64 parameters, instead of casting to u64 first (`266822a`).

src/tables/u64_set.rs
```diff
@@ -47,12 +47,20 @@ impl U64Set {
...
+    /// final merge -- see [crate::pipeline::EXTERNAL_SORT_CHUNK_BYTES]).
+    /// Callers that run several external sorts concurrently (e.g. from
+    /// sibling threads in the same `thread::scope`) should divide their
+    /// share of the budget accordingly -- this function has no way to
+    /// know how many others are running alongside it.
     pub fn create(
         elements: impl Iterator<Item = u64>,
         workdir: &Path,
         out: &Path,
+        chunk_bytes: u64,
```
brawer: Make chunk_bytes also an usize (in all places being chnanged), again for readability. We'll never run on 32-bit machines.

brawer: Fixed -- chunk_bytes is now usize on all six table constructors (BlobTable, GeometryTable, CoordTable, GraphTable, U64Set, StringCounts), matching EXTERNAL_SORT_CHUNK_BYTES's own type. The two call sites that build a MemoryLimitedBufferBuilder directly still cast to u64 right at that call, since that's what the ext_sort crate's API itself requires (`4b4b88a`).

src/tables/string_pool.rs
```diff
@@ -542,11 +543,17 @@ impl Writer {
         let mut hash_values_writer =
             BufWriter::with_capacity(32 * 1024, File::create(&hash_values_path)?);
 
+        // StringPool::create has exactly one caller in the pipeline
+        // (assemble_strings), which never runs another external sort
+        // concurrently, so this gets the full chunk-size budget; see
+        // crate::pipeline::EXTERNAL_SORT_CHUNK_BYTES. Entries are
+        // fixed-size, so the budget is converted to an item count rather
+        // than measured via MemoryLimitedBufferBuilder.
         let sorter: ExternalSorter<(u32, usize), std::io::Error, LimitedBufferBuilder> =
```
brawer: Can we define a local type `Item` (or whatever), and use it to both set up the ExternalSorter, and to compute the sort buffer size? I'm worried that, in a future refactoring, someone would overlook the repeated item type, so the type would not match anymore the buffer size computation. As it's written now, the compiler would not flag this as an error, so it would go unnoticed. (Likewise in the other places).

brawer: Fixed -- introduced a local `type Item = ...;` alias in string_pool.rs and reused it for both the `ExternalSorter<Item, ...>` and `size_of::<Item>()`. Applied the same pattern to the other fixed-size (LimitedBufferBuilder-based) sorts touched by this PR too: coord_table.rs ((u64, u64)), graph.rs (Edge), and u64_set.rs (u64) (`ef54dac`). The MemoryLimitedBufferBuilder-based ones (blob_table.rs, string_counts.rs, and the hardcoded sites in conflate/edits/atp/assemble_strings) don't call size_of at all, so there's nothing to drift there. feature_index.rs's Pass1Item/Pass2Item from #656 already followed this pattern.

brawer (commented):
LGTM

Issue #666, opened 2026-08-13
Expose osm.timestamp and atp.timestamp in conflated.parquet


brawer:
`conflated.parquet` doesn't currently carry a last-edit/last-seen timestamp for either side, even though the underlying data for both is already available close to where rows get built -- just not plumbed through to the schema.

## OSM side

\`tables::Feature\` (\`src/tables/feature.proto\`) already has a \`timestamp\` field: "Unix timestamp (seconds since 1970-01-01 UTC) of the last edit. 0 if unknown." It's populated by \`assemble_feature()\` in \`pipeline/osm/assemble.rs\` from the PBF's own per-element metadata. Once conflation reads OSM features via \`OsmFeatureIndex\`/\`OsmCandidate\` (rather than \`Place\`, which has no timestamp field), \`feature.timestamp\` is just sitting there on every matched candidate.

## ATP side

Also closer than it looks: \`atp::make_place(geojson: &str, _timestamp: time::UtcDateTime) -> Option<Place>\` (\`src/atp/mod.rs\`) already receives a per-feature timestamp as a parameter -- it's just discarded today (leading underscore, unused). \`Place\` itself has no timestamp field to put it in yet.

## Why not now

Not related to the OSM-pipeline refactoring currently in flight, so deliberately not blocking or blocked by it -- but sequencing makes sense to wait for:
- Once the OSM side conflates via \`OsmFeatureIndex\`/\`Feature\` instead of \`Place\`, adding \`osm.timestamp\` to \`ParquetRow\`/the conflate writer schema should be a small, localized change (the value is already decoded and in hand at \`ParquetRow::from_conflated\`'s call site).
- \`atp.timestamp\` needs a field added to \`Place\` (or whatever replaces it, if/when \`Place\` itself gets reworked/retired for the ATP side too) -- doing that before or after such a change affects how much churn it causes, so worth waiting to see which comes first.

## Scope

Add \`timestamp\` (nullable, Unix seconds) to both the \`osm\` and \`atp\` struct columns in \`conflated.parquet\`'s Arrow/Parquet schema (\`pipeline/conflate/writer.rs\`), sourced from \`Feature.timestamp\` and the currently-discarded \`_timestamp\` parameter in \`atp::make_place\`, respectively.

brawer:
[...]

PR #686, opened 2026-08-17
Populate Place.fetched from ATP's spider:collection_time

brawer:
First of several PRs preparing `conflated.parquet`'s schema for v0.7.0
(see discussion on #666, #645) -- scoped down to just the ATP
fetch-timestamp plumbing, so it's reviewable on its own before the
larger `conflated.parquet` rewrite (osm.modified, way_members/
relation_members, non-nullable tag values) lands separately.

## What

`process_geojson` already parsed `spider:collection_time` out of each
FeatureCollection's `dataset_attributes` and passed it down to
`make_place()` -- but `make_place()` declared the parameter as
`_timestamp` and silently discarded it. This wires it through into a
new `Place.fetched` field, round-tripped through `alltheplaces.parquet`
(a new `fetched` `Int64` column, Unix seconds -- that file is purely an
internal scratch format, so no need for a fancy logical type there).

## The DeepSizeOf wrinkle

`Place` derives `DeepSizeOf` (needed for `MemoryLimitedBufferBuilder`'s
external-sort accounting), which `time::UtcDateTime` doesn't implement
-- and can't be made to from here either, since both the trait and the
type are foreign to this crate, so Rust's orphan rule blocks a direct
`impl`. Added `crate::utils::UtcTimestamp`, a small local wrapper that:
- implements `DeepSizeOf` trivially (`UtcDateTime` has no heap
  allocation, so there's nothing to add beyond the wrapper's own
  shallow size),
- (de)serializes as a plain `i64` Unix-seconds integer rather than an
  RFC 3339 string, since `Place`'s `Serialize`/`Deserialize` impl is on
  a hot path (MessagePack, once per spilled chunk of every external
  sort of the ATP dataset) -- unlike `crate::utils::rfc3339`, which
  exists for `provenance.rs`'s one-off JSON output and would cost real
  formatting/parsing time and extra spilled bytes here for no benefit.

## Fixture

Regenerated `tests/test_data/alltheplaces.parquet` (a static fixture)
with the new column -- verified identical content otherwise via DuckDB
before and after, using `tests/test_data/alltheplaces.meta.json`'s
`start_time` as the synthetic `fetched` value.

## Scope

Internal plumbing only -- `conflated.parquet` doesn't expose
`atp.fetched` yet. That (plus `osm.modified`, `way_members`/
`relation_members`, and non-nullable tag values) is a separate,
larger PR.

## Testing

`cargo build --locked --all-targets`, `cargo test --locked` (238 unit +
4 integration tests), `cargo fmt --check`, `cargo clippy --locked
--all-targets -- -D warnings` all clean.


brawer (commented):
LGTM

Issue #663, opened 2026-08-13
GeoParquet: Make conflated.parquet a fully conformant, verified GeoParquet 2.0 file


brawer:
### Summary

`conflated.parquet` currently nests `geometry` (a Parquet `GEOGRAPHY`) inside both the `osm` and `atp` structs. I'd like to propose moving these to top-level columns — `osm_geometry` and `atp_geometry` — to bring the file into line with GeoParquet 2.0's requirement that geometry columns live at the schema root, and to get reliable spatial pruning/tooling support in return.

**Update — scope grew while implementing this:** moving the two columns to the schema root turned out to be necessary but not sufficient. This repo wasn't writing GeoParquet's file-level `geo` Parquet key-value metadata (`version`/`primary_column`/`columns`) at all — required for a file to be conformant GeoParquet 2.0, not just "geometry happens to sit at the top level". So the actual scope here is: make `conflated.parquet` a genuinely conformant, *verified* GeoParquet 2.0 file, not just relocate two columns. Concretely, on top of the column move:

- Write the file-level `geo` metadata: `version: "2.0.0"` (GeoParquet 2.0 is still at `v2.0.0-rc.1`, the newest spec available; its on-disk format, including this fixed version string, isn't expected to change again before the final release), `primary_column: "osm_geometry"` (OpenStreetMap is the dataset a reviewer is actually meant to correct), and per-column `encoding`/`geometry_types`/`edges`.
- Compute `geometry_types` for real per column (`["Point"]`, `["Polygon"]`, etc.), rather than leaving it an empty "unknown" array — read off each row's WKB header (endianness + type code) without decoding coordinates.
- Make the columns' native Parquet `GEOGRAPHY` logical type's CRS (`OGC:CRS84`) agree with the `geo` metadata's default CRS, instead of leaving a latent `EPSG:4326`-vs-`OGC:CRS84` mismatch between the two.
- Actually **verify** conformance against a real validator — [`gpio`](https://geoparquet.io/guide/check/) (the `geoparquet-io` CLI) — rather than just reading the spec and hoping the implementation matches it.

Done in #688.

### Current schema (simplified)

```
osm: struct
  ├─ tags: map<string, string>
  ├─ timestamp: timestamp
  ├─ changeset: int64
  ├─ way_members: list<int64>
  └─ geometry: GEOGRAPHY   ← nested
atp: struct
  ├─ tags: map<string, string>
  ├─ timestamp: timestamp
  ├─ spider: string
[...]

PR #687, opened 2026-08-17
Rewrite conflated.parquet's schema: modified/fetched, members, non-null tags

### LINKED ISSUE #645: Write docs/CONFLATED_OUTPUT.md describing the conflated.parquet schema

brawer:
Closes #645. Related to #666, #663 (not fully closing #663 -- see below).

Bundles several schema-breaking changes into `conflated.parquet` in one
go, since the file has never been published -- there are no external
clients to break yet, so this is the moment to get the shape right
without needing a major-version bump or a compatibility note. Follows
#686 (`Place.fetched`/`atp.fetched` internal plumbing), now wired all
the way through to the actual output.

## Schema changes

- `osm.changeset`/`osm.version` move into a new `osm.modified` struct
  alongside a new `osm.modified.timestamp` (OSM's own edit timestamp --
  already available on `tables::Feature`, just never exposed before).
- New `atp.fetched` struct (`spider` + `timestamp`), replacing the old
  flat `atp.spider` field -- mirrors `osm.modified`: both answer "who
  produced this side of the row, and when".
- New `osm.way_members` (`list<u64>`, present only for ways) and
  `osm.relation_members` (`list<struct{type,id,role}>`, present only for
  relations) -- both already computed by `assemble.rs` onto
  `tables::Feature`, just not wired through to output before.
- `atp.tags`/`osm.tags` map values are now non-nullable, matching how
  `alltheplaces.parquet`'s own `tags` column already worked; nothing
  ever actually wrote a null value here, so this only tightens the
  schema to match reality.

## The timestamp-unit wrinkle

Timestamps use Arrow's `Timestamp(Millisecond, UTC)`, not `Second`:
Parquet's `TIMESTAMP` logical type has no "seconds" unit at all
(confirmed in the `parquet` crate's own `arrow_to_parquet_type` --
`Timestamp(Second, _)` silently produces a bare, unannotated `INT64`
column, not a real logical timestamp -- no error, just silent
degradation). Verified via DuckDB before and after: with `Second`, the
column read back as plain `BIGINT`; with `Millisecond`, DuckDB resolves
it natively as `TIMESTAMP WITH TIME ZONE`. Source data only ever has
whole-second precision, so this loses nothing.

## Other changes

- `pipeline::edits.rs`'s `extract_conflated_row` updated for the nested
  `osm.modified` struct (new `get_child_struct` helper, mirroring the
  one already in this file's sibling reader code).
- `tests/integration_test.rs`'s `assert_conflated_parquet` extended to
  pin concrete values for all the new fields (changeset/version/
  timestamp, `way_members` length, ATP spider/fetched) -- cross-checked
  against the fixture data via DuckDB directly, not just re-derived
  from this same code.
- Added `docs/CONFLATED_OUTPUT.md` (#645): full schema reference, data
  provenance (CycloneDX BOM) pointer, encoding notes, and a DuckDB
  quick-look example -- linked from `docs/README.md`. Un-stubbed the
  `TODO(#645)` comment in `provenance.rs` that pointed at this doc
  before it existed.

## Not in scope here

#663 (move `geometry` to top-level columns for strict GeoParquet
compliance) is intentionally *not* done in this PR -- flagged in the new
doc as a known gap, still tracked separately.

## Testing

`cargo build --locked --all-targets`, `cargo test --locked` (238 unit +
4 integration tests, including the full `test_pipeline` end-to-end run
with pinned assertions on every new field), `cargo fmt --check`,
`cargo clippy --locked --all-targets -- -D warnings` all clean. Also
ran the real pipeline against the fixtures manually and inspected the
output with DuckDB (schema, logical types, and actual values) before
writing the automated assertions.


docs/CONFLATED_OUTPUT.md
```diff
@@ -0,0 +1,126 @@
+# `conflated.parquet`
+
```
brawer: Let's move this into docs/outputs/CONFLATED_PARQUET.md, since we'll have more outputs, and this link should be stable (since its location is baked into the file metadata). Add a docs/outputs/README.md, stating that this directory documents our output files intended for public consumption. Link both from the top-level README and from docs/README to docs/outputs(/README).

docs/CONFLATED_OUTPUT.md
```diff
@@ -0,0 +1,126 @@
...
+This is the pipeline’s public output: for every feature AllThePlaces
```
brawer: Other than for the other documentation we have so far, this doc (and other future docs in docs/output) is actually not intended for programmers who join the project. Rather, the typical reader is a GIS specialist with little to no programming experience — they can use QGIS and DuckDB, maybe a bit of Python scripting, but the may never have heard of Rust (or care about it). So, don't link to Rust code from here, just write it so the text makes sense to someone who just wants to read the data into a GIS tool, or hack together a little throwaway Python script. Also, no need to document or link to how we compute the matches; we'll eventually have a separate design document for that — but the people who consume our data won't really care.

docs/CONFLATED_OUTPUT.md
```diff
@@ -0,0 +1,126 @@
...
+  tags        MAP<UTF8, UTF8>                non-null map, non-null keys and values
+  modified                                   STRUCT, non-null
+    timestamp TIMESTAMP(ms, UTC)             non-null
+    changeset UINT64                         non-null
+    version   UINT32                         non-null
+  way_members LIST<UINT64>                   nullable — present only when type = "way"
+  relation_members LIST<STRUCT>              nullable — present only when type = "relation"
+    type      UTF8                           non-null ("node" | "way" | "relation")
+    id        UINT64                         non-null
+    role      UTF8                           non-null (empty string if OSM gave the member no role)
```
brawer: Hm, let's make this nullable, and pass null if OSM have no role. No need to change our internal representations (unless it's trivial, then let's change those too); but in the public output schema, this should be null if OSM gives no role to a member.

docs/CONFLATED_OUTPUT.md
```diff
@@ -0,0 +1,126 @@
...
+  geometry    BINARY (WKB)                   non-null
+```
+
+A few things worth calling out that aren’t obvious from the shape alone:
```
brawer: "from the shape alone" -> "from the above" — in a GIS context, "shape" sounds like geometry

docs/CONFLATED_OUTPUT.md
```diff
@@ -0,0 +1,126 @@
...
+
+- **`atp` is nullable, but every row emitted today has it set.** The
+  slot exists for a possible future “this OSM feature exists, but ATP
+  doesn’t track it” row (see
+  [alltheplaces/osm-diffs#682](https://github.com/alltheplaces/osm-diffs/issues/682)),
+  not because `conflate()` produces one today.
+- **`osm` is `NULL` exactly when nothing in OpenStreetMap matched.**
+  That’s the normal, common case — most ATP features don’t have an OSM
+  counterpart yet — not a data-quality problem.
```
brawer: Not sure if it's true that most ATP features have no counterpart yet. Omit this quantitative claim, and just say it can happen.

docs/CONFLATED_OUTPUT.md
```diff
@@ -0,0 +1,126 @@
...
+  most recent edit — OSM’s edit history itself isn’t exposed, just its
+  current state).
+- **`osm.way_members`/`osm.relation_members` are `NULL`, not an empty
+  list, when they don’t apply** — a `NULL` `way_members` means “this
+  isn’t a way”, not “this way has no members”. Only one of the two is
+  ever non-`NULL` for a given row, matching `osm.type`.
+  `way_members` is a way’s node references in OSM’s declared order,
+  as plain node IDs. `relation_members` is a relation’s members in OSM’s
+  declared order; each member’s own `id` uses the same
+  `osm_id * 10 + {1,2,3}` encoding internal to this pipeline being
```
brawer: This formula is internal to our pipeline, don't mention it in this doc. I think this can be omitted.

docs/CONFLATED_OUTPUT.md
```diff
@@ -0,0 +1,126 @@
...
+  already decoded away by the time it reaches this column — what you
+  get is a plain OSM ID plus a `type` telling you which of node/way/
+  relation it refers to.
+- **Timestamps are milliseconds, not seconds** — Parquet’s `TIMESTAMP`
+  logical type has no “seconds” unit (only millis/micros/nanos), so
```
brawer: Actually, the ATP collection time is recorded in nanoseconds. We don't need to emit nanoseconds here, but in its current state the comment is likely wrong: From ATP, we should see actual milliseconds. In anycase, since we're using a native Parquet type with clear granularity (ms), we probably don't need to document it here. But please double-check that we're actually passing millisecond accuracy from ATP collection time up to the atp timestamp in conflated.parquet; if it somewhere gets rounded to seconds, we should fix it. — Also, don't mentinon "spider:collection_time" here, since this will probably confuse GIS specialists who know nothing about the details of AllThePlaces.

docs/CONFLATED_OUTPUT.md
```diff
@@ -0,0 +1,126 @@
...
+  seconds-precision source data (both `spider:collection_time` and
+  OSM’s edit timestamps only ever carry whole seconds) is stored as
+  milliseconds throughout, always exact multiples of 1000.
+- **`geometry` is real OGC Simple Features WKB**, not a centroid or a
```
brawer: Try to find a background page that explains OGC Simple Features, and link to it. The Wikipedia page is difficult to read though, hopefully you can find something better. — Not sure if should we really talk about WKB here, it seems like a technical encoding detail. Also, remove "not a centroid or a synthetic point", just say that atp.geometry iss taken from the upstream source — nearly always a point, but we do ingest a few sources that provide non-point geometry on their features.

docs/CONFLATED_OUTPUT.md
```diff
@@ -0,0 +1,126 @@
...
+  a `LineString` for a way that isn’t one, a `Point` for a node). Not
+  strict [GeoParquet](https://geoparquet.org/): there’s no top-level
+  `geo` metadata key, since the geometry columns are nested inside
+  `atp`/`osm` rather than top-level (tracked in
+  [alltheplaces/osm-diffs#663](https://github.com/alltheplaces/osm-diffs/issues/663)).
+  Typing rides on Arrow’s own WKB extension-type mechanism instead
+  (`EPSG:4326`, spherical edges) — tools that understand Arrow
+  extension types (DuckDB, GeoArrow-aware readers) pick it up
+  automatically; a generic Parquet reader just sees `BINARY`.
+- **No internal sort key is exposed.** Rows are written in ascending S2
```
brawer: Shorten this. Maybe something like: For better compression and spatial query performance, rows are sorted along a Hilbert curve. At the moment, we compute the S2 cell ID of the geometry's centroid point, but we don't want to commit to a specific way of Hilbert sorting, so we don't expose the sort key. (But feel free to change my words so it reads better).

docs/CONFLATED_OUTPUT.md
```diff
@@ -0,0 +1,126 @@
...
+## Data provenance
+
+Every file carries a [CycloneDX](https://cyclonedx.org/) JSON document
+in its Parquet key-value metadata, under the key `org.cyclonedx.bom` —
+answering “which version of this pipeline, using exactly what input,
+and at what time, produced this file”. It records the AllThePlaces
+dump and OpenStreetMap planet snapshot consumed (with their own
+upstream identifiers/timestamps/hashes where available), the pipeline
+version and its own supply-chain identity, and the run’s
+start/end time. See [`src/provenance.rs`](../src/provenance.rs) for
```
brawer: Remove the sentence starting with "See" until the end of the paragraph, it's too much detail for the audience. (GIS specialists don't know about supply chain security, and probably couldn't care less). Instead, how about a DuckDB command that illustrates how to find the ATP run, OSM replication timestamp, and our pipeline's version tag, from a `conflated.parquet` file? Maybe say that we use CycloneDX because it's an industry-format for exactly this kind of info, but don't talk about releases or containers; GIS people don't know what these are, and it's not really relevant when consuming this file.

docs/CONFLATED_OUTPUT.md
```diff
@@ -0,0 +1,126 @@
...
+exactly what’s in it, and
+[`SUPPLY_CHAIN_SECURITY.md`](SUPPLY_CHAIN_SECURITY.md) for the
+concepts behind it (that document is about the *release container
+image’s* own SBOM/attestations — a related but distinct thing from
+this per-file data provenance).
+
+## Encoding
```
brawer: Omit this paragraph, too much technical detail for the audience.

docs/CONFLATED_OUTPUT.md
```diff
@@ -0,0 +1,126 @@
...
+
+```sql
+-- Real OSM shapes, not synthetic points:
+SELECT osm.type, osm.id, atp.fetched.spider, ST_GeometryType(osm.geometry)
+FROM read_parquet('conflated.parquet')
+WHERE osm.id IS NOT NULL
+LIMIT 10;
+```
+
+DuckDB (with the `spatial` extension loaded) reads `atp.geometry`/
```
brawer: Maybe show the command line to invoke DuckDB? — As of August 2026, is it likely that someone has DuckDB without spatial support? If you really think so, let's show how to check for the spatial extension; but honestly, I think we can omit this phrase, especially since the audience is GIS peple.

docs/CONFLATED_OUTPUT.md
```diff
@@ -0,0 +1,126 @@
...
+`osm.geometry` as its native `GEOMETRY` type and `atp.fetched.timestamp`/
```
brawer: Mabe we can "show, not tell" here? Ie., just show (with LIMIT 1 or 2) what users see for the command, then we don't need to explain it.

src/provenance.rs
brawer: Let's make this link machine-readable, as a second entry in externalReferences. Please research what's the most appropriate way to indicate this in CycloneDX 1.7, and use an appropriate tag.

docs/CONFLATED_OUTPUT.md
brawer: Done -- moved to docs/outputs/CONFLATED_PARQUET.md and added docs/outputs/README.md introducing the directory, linked from the top-level README and docs/README (`945a326`).

docs/CONFLATED_OUTPUT.md
brawer: Rewrote for that audience: no links into the Rust source, no mention of how matching/conflation works (that's a separate future design doc, as you say), no internal implementation details. (`945a326`)

docs/CONFLATED_OUTPUT.md
brawer: Done -- role is now nullable in the output schema (null when OSM gives a member no role), verified via DuckDB against a real run (repetition_type: OPTIONAL). Internal representation unchanged, per your suggestion -- still a plain String, interpreted at write time (empty string -> null). (`945a326`)

docs/CONFLATED_OUTPUT.md
brawer: Fixed (`945a326`).

docs/CONFLATED_OUTPUT.md
brawer: Fixed -- now just says it can happen, no quantitative claim (`945a326`).

docs/CONFLATED_OUTPUT.md
brawer: Removed (`945a326`).

docs/CONFLATED_OUTPUT.md
brawer: Linked PostGIS's own geometries introduction instead of Wikipedia -- clear and practitioner-friendly, verified. Removed the WKB mention and the 'not a centroid/synthetic point' phrasing; now just says atp.geometry comes from AllThePlaces' own scrape, nearly always a point but a handful of sources give lines/polygons (`945a326`).

docs/CONFLATED_OUTPUT.md
brawer: Used almost exactly your wording (`945a326`).

docs/CONFLATED_OUTPUT.md
brawer: Good catch -- there really was a bug: UtcTimestamp's own serialize/deserialize (used on every Place's round trip through the ATP external sort) and alltheplaces.parquet's internal fetched column both went through unix_timestamp(), which discards anything below whole seconds. Denner's actual source value is ...952.804399 (microseconds, not nanoseconds, but same problem) -- it really was getting silently rounded down before ever reaching conflated.parquet. Fixed both spots to preserve millisecond precision (see UtcTimestamp::unix_timestamp_millis, computed from unix_timestamp_nanos), regenerated the fixture, and added a regression test pinning the exact value. Confirmed end-to-end via the integration test: Denner's atp.fetched.timestamp is now 1780209952804 (not ...952000). Also dropped the mention of spider:collection_time from the doc -- the schema listing already says milliseconds, no need to repeat it or name the ATP-internal source field (commit 945a326).

docs/CONFLATED_OUTPUT.md
brawer: Rewrote the whole section: dropped the supply-chain-security/release/container detail, added a DuckDB example (with real output) that pulls the pipeline version, AllThePlaces run, and OpenStreetMap snapshot out of the embedded BOM, described CycloneDX as an established format for this kind of provenance record, and mentioned that licensing info for both inputs and the output is included too (commit 945a326).

docs/CONFLATED_OUTPUT.md
brawer: Removed (commit 945a326).

docs/CONFLATED_OUTPUT.md
brawer: Showed the duckdb CLI invocation for both examples now. Kept the INSTALL/LOAD spatial lines (the quick-look query needs ST_GeometryType), but just as part of the shown command, no prose explaining why (commit 945a326).

docs/CONFLATED_OUTPUT.md
brawer: Done -- both examples now show the actual command and real, verified output (LIMIT 3) instead of describing it in prose (commit 945a326).

src/provenance.rs
brawer: Looked up the real CycloneDX 1.7 schema (bom-1.7.schema.json) for externalReferenceType. Best fit is "evidence" ("Information used to substantiate a claim") -- this link is exactly evidence substantiating the claim that using CC0-licensed ATP data in this ODbL-licensed pipeline is clear. Added it as a second entry in atp_component()s externalReferences, pulled the URL into a named constant with its own doc comment explaining why (commit 945a326).

docs/outputs/CONFLATED_PARQUET.md
```diff
@@ -0,0 +1,124 @@
...
+    id        UINT64                         non-null
+    role      UTF8                           nullable — null if OSM gives the member no role
+  geometry    BINARY (WKB)                   non-null
+```
+
+A few things worth calling out that aren’t obvious from the above:
+
+- **`osm` is `null` whenever nothing in OpenStreetMap matched.** That
+  can happen for perfectly ordinary reasons — it doesn’t mean anything
+  is wrong with the row.
```
brawer: Change "is wrong with the row" to "is wrong with the conflation pipeline; rather, it is a signals that some data may be missing from OSM." (Or similar; feel free to change and simplify my wording).

docs/outputs/CONFLATED_PARQUET.md
```diff
@@ -0,0 +1,124 @@
...
+- **`atp.fetched` and `osm.modified` mirror each other on purpose**:
+  both answer “who produced this side of the row, and when” —
+  `atp.fetched` for AllThePlaces’ own scrape, `osm.modified` for
+  OpenStreetMap’s own edit metadata (the timestamp, changeset, and
+  version number of that feature’s most recent edit — not its full
```
brawer: edit. We don’t preserve the full edit history of OpenStreetMap here, only the current state at the time of the database snapshot.

docs/outputs/CONFLATED_PARQUET.md
```diff
@@ -0,0 +1,124 @@
...
+
+## Geometry
+
+`atp.geometry`/`osm.geometry` are standard [OGC Simple
+Features](https://postgis.net/workshops/postgis-intro/geometries.html)
+geometries — the same model QGIS, PostGIS, and most other GIS tools
+already use. `atp.geometry` is whatever AllThePlaces’ own scrape
+provides for that feature — nearly always a point, though a handful of
+sources provide lines or polygons instead. `osm.geometry` is the
+matched OpenStreetMap feature’s actual shape: a polygon for an area, a
```
brawer: Remove "actual" from "actual shape", and say (pehaps as another sentence at the end of this paragraph, it would fit nicely there, or maybe in a separate paragraph) that osm.geometry is not always exactly as in OSM, because (a) We try to repair certain errors (such as lines that self-intersect), and (b) if the geometry would become very large, we simplify its shape to limit its size (number of coordinates) small, so that downstream users can eg. compute a GeoJSON for display in a web browser. For simplification, we use a cartographic algorithm (currently, a topology-preserving variant of the [Visvalingam-Whyatt algorithm](https://en.wikipedia.org/wiki/Visvalingam%E2%80%93Whyatt_algorithm), but this may change). — In your blurb, don't refer to the Rust code in the geometry module (because the audience for this doc are GIS people, not programmers), but it's important to state that we do _not_ always preserve geometry as in OSM. So, osm.geometry is something they can use for display, eg to mark the affected OSM feature on a map. But if they want to generate edits on OSM, they should use way_members and relation_members.

docs/outputs/CONFLATED_PARQUET.md
```diff
@@ -0,0 +1,124 @@
...
+line for a way that isn’t an area, a point for a node.
+
+Rows are sorted along a Hilbert curve, for better compression and
```
brawer: Make "Hilbert curve" a link? The Wikipedia article is rather mathematical, hopefully you can find a page that is easier to read for a GIS audience. (Also search for "space-filling curve", "google s2 cell geometry", etc.)

docs/outputs/CONFLATED_PARQUET.md
```diff
@@ -0,0 +1,124 @@
...
+faster spatial queries. At the moment, we compute this from the S2
+cell ID of each geometry’s centroid, but we don’t want to commit to
+that particular way of doing the sort, so the sort key itself isn’t
+exposed as a column.
+
+## Data provenance
```
brawer: I'd suggest moving this section (including the duckdb snippet) to the end. Metadata is less important than the actual content.

docs/outputs/CONFLATED_PARQUET.md
```diff
@@ -0,0 +1,124 @@
...
+
+Every file carries a [CycloneDX](https://cyclonedx.org/) document
+embedded in its own metadata — CycloneDX being an established,
+widely-used format for exactly this kind of “where did this data come
```
brawer: Substitute "widely-used" by [standardized](https://ecma-international.org/wp-content/uploads/ECMA-424_2nd_edition_december_2025.pdf)

docs/outputs/CONFLATED_PARQUET.md
brawer: Used close to your wording -- "is wrong with the conflation pipeline; rather, it's a signal that the feature may be missing from OpenStreetMap" (commit 1c47ffb).

docs/outputs/CONFLATED_PARQUET.md
brawer: Done -- spelled it out explicitly: "We do not preserve OpenStreetMap's full edit history here, only its current state as of the database snapshot this file was built from" (commit 1c47ffb).

docs/outputs/CONFLATED_PARQUET.md
brawer: Done -- dropped "actual", and added a paragraph explaining that osm.geometry is not always byte-for-byte OSM's data: we repair self-intersections automatically, and simplify (topology-preserving Visvalingam-Whyatt, verified this is really what src/geometry/ does) when a shape would otherwise get very large. Says explicitly: fine to display, use way_members/relation_members if you need to actually generate an edit. No Rust references (commit 23e9042).

docs/outputs/CONFLATED_PARQUET.md
brawer: Linked to a Towards Data Science article on spatial-index space-filling curves instead -- verified it explains Hilbert/Z-order curves without heavy math notation, unlike Wikipedia's treatment (commit 23e9042).

docs/outputs/CONFLATED_PARQUET.md
brawer: Moved -- Data provenance now comes after A quick look (commit 23e9042).

docs/outputs/CONFLATED_PARQUET.md
brawer: Done (commit 23e9042).

PR #688, opened 2026-08-18
conflated.parquet: make it valid GeoParquet 2.0-rc.1

### LINKED ISSUE #663: GeoParquet: Make conflated.parquet a fully conformant, verified GeoParquet 2.0 file

brawer:
Fixes #663.

## What changed

- `atp.geometry`/`osm.geometry` move out of the `atp`/`osm` structs into top-level `atp_geometry`/`osm_geometry` columns. GeoParquet requires geometry columns to live at the schema root, not nested inside structs — this was the core ask in #663.
- Adds the file-level `geo` key-value metadata GeoParquet 2.0 itself requires (this repo didn't write it at all before, for any output file): `version`, `primary_column` (`osm_geometry` — OpenStreetMap is the dataset a reviewer is actually meant to correct), and per-column `encoding`/`geometry_types`/`edges`.
- `geometry_types` is computed for real, not left as an empty "unknown" array. New `src/geometry/wkb_type.rs` adds a `WkbGeometryType` enum and a `wkb_geometry_type()` reader that peeks a WKB buffer's header (byte order + type code) without decoding coordinates. `GeometryTally` (in `writer.rs`) tracks a real per-type count for each column across the whole file, plus the single largest geometry seen (bytes + a lazily-computed label — OSM `type/id`, ATP spider — only formatted when a row actually becomes the new largest) — both the counts and the largest-geometry info are logged at INFO in `close()`, one record each for `atp_geometry`/`osm_geometry`.
- `crs` is deliberately omitted from the `geo` metadata on both columns: it defaults to `OGC:CRS84` per spec, which is also what the columns' native Parquet `GEOGRAPHY` logical type now requests (switched from `EPSG:4326` to `OGC:CRS84`), so the two stay consistent without spelling out a PROJJSON object.
- `edits.rs`'s `suggest_edits` also reads `conflated.parquet` and needed the same top-level-column fix for `osm_geometry`.
- `docs/outputs/CONFLATED_PARQUET.md` updated: schema, the "quick look" query, a new intro paragraph stating this is a GeoParquet 2.0-rc.1 file with a link to the spec, and some editorial fixes from review (GeoJSON/GeoPackage mentioned as formats not tools, geometry simplification reframed around the downstream client's benefit).
- `PROVENANCE_KEY` and `GEO_METADATA_KEY` are now both set the same way, via `append_key_value_metadata` in `close()`, rather than one going through `WriterProperties` up front and the other appended late.

## Why 2.0-rc.1, not 1.1

GeoParquet 2.0 hasn't had a final release yet (`v2.0.0-rc.1`, July 2026, is the newest tag), but its on-disk format — including the file-level metadata's `version` field, which is a fixed `"2.0.0"` string — isn't expected to change again before the final release. Targeting 1.1 would mean re-doing this work again once 2.0 actually ships, and 1.1's classic `BYTE_ARRAY`-only columns wouldn't get us the native `GEOMETRY`/`GEOGRAPHY` Parquet logical type this repo already uses (`parquet-geospatial`), so 2.0-rc.1 is the right target now.

## Verification

Ran the pipeline against the test fixtures to produce a real `conflated.parquet`, then:

- [`gpio`](https://geoparquet.io/guide/check/) (the `geoparquet-io` CLI validator): `gpio check spec` and `gpio check all` both report **49 passed, 0 warnings, 0 failed**.
- Cross-checked the raw Parquet metadata directly via `parquet-tools meta` and DuckDB's `parquet_schema()`/`DESCRIBE`, confirming the native logical type, geospatial statistics, and the embedded `geo` JSON all match what's documented above.
- Checked `pipeline.log` directly for the new `conflate.write: atp_geometry geometry types` / `conflate.write: osm_geometry geometry types` records.
- `cargo test`, `cargo clippy --all-targets`, `cargo fmt --check` all clean.

One nuance investigated during review, not a bug: `parquet-tools meta`/DuckDB show the native Parquet `GEOGRAPHY` logical type's own `crs`/`algorithm` sub-fields as unset. Built an isolated repro against plain `arrow`/`parquet`/`parquet-geospatial` (no dependency on this pipeline) and traced it into `parquet`'s own source (`logical_type_for_binary`, `parquet` 59.2): a lon/lat CRS and `Spherical` edges are deliberately canonicalized to `None` there, because those are themselves the Parquet spec's own defaults for `GEOGRAPHY` — the same omit-if-default encoding GeoParquet's own `geo` metadata uses for `crs`. So this is the correct, intentional encoding of exactly the `OGC:CRS84`/spherical values requested, not data loss — no bug report needed.

### DIFF
--- a/docs/outputs/CONFLATED_PARQUET.md
+++ b/docs/outputs/CONFLATED_PARQUET.md
@@ -7,6 +7,13 @@ pairing the AllThePlaces feature with whichever OpenStreetMap feature
 it was matched to — or with no OpenStreetMap match at all, if none was
 found.
 
+This is a [GeoParquet 2.0-rc.1](https://github.com/opengeospatial/geoparquet/blob/v2.0.0-rc.1/format-specs/geoparquet.md)
+file (2.0 hasn’t had its final release yet, but this is the newest spec
+available, and the on-disk format isn’t expected to change again before
+it does). In practice that means any GeoParquet-aware tool — QGIS,
+kepler.gl, GeoPandas — can open it directly and will find both
+geometry columns on its own, without being told where to look.
+
 ## Schema
 
 ```text
@@ -15,7 +22,7 @@ atp                                          STRUCT, nullable
   fetched                                    STRUCT, non-null
     timestamp TIMESTAMP(ms, UTC)             non-null
     spider    UTF8                           non-null
-  geometry    BINARY (WKB)                   non-null
+atp_geometry  BINARY (WKB)                   nullable — null exactly when atp is null
 
 osm                                          STRUCT, nullable
   type        UTF8                           non-null ("node" | "way" | "relation")
@@ -30,7 +37,7 @@ osm                                          STRUCT, nullable
     type      UTF8                           non-null ("node" | "way" | "relation")
     id        UINT64                         non-null
     role      UTF8                           nullable — null if OSM gives the member no role
-  geometry    BINARY (WKB)                   non-null
+osm_geometry  BINARY (WKB)                   nullable — null exactly when osm is null
 ```
 
 A few things worth calling out that aren’t obvious from the above:
@@ -39,6 +46,14 @@ A few things worth calling out that aren’t obvious from the above:
   can happen for perfectly ordinary reasons — it doesn’t mean anything
   is wrong with the conflation pipeline; rather, it’s a signal that
   the feature may be missing from OpenStreetMap.
+- **`atp_geometry`/`osm_geometry` sit outside the `atp`/`osm` structs**,
+  at the top level of the file, rather than next to the rest of each
+  side’s data — GeoParquet requires geometry columns to live at the
+  schema root, which is what lets a generic GIS tool discover them on
+  its own instead of needing to be told where to look.
+  `osm_geometry` is the file’s designated primary geometry column,
+  since OpenStreetMap is the dataset a reviewer is actually meant to
+  correct.
 - **`atp.fetched` and `osm.modified` mirror each other on purpose**:
   both answer “who produced this side of the row, and when” —
   `atp.fetched` for AllThePlaces’ own scrape, `osm.modified` for
@@ -57,23 +72,26 @@ A few things worth calling out that aren’t obvious from the above:
 
 ## Geometry
 
-`atp.geometry`/`osm.geometry` are standard [OGC Simple
+`atp_geometry`/`osm_geometry` are standard [OGC Simple
 Features](https://postgis.net/workshops/postgis-intro/geometries.html)
 geometries — the same model QGIS, PostGIS, and most other GIS tools
-already use. `atp.geometry` is whatever AllThePlaces’ own scrape
-provides for that feature — nearly always a point, though a handful of
-sources provide lines or polygons instead. `osm.geometry` is the
-matched OpenStreetMap feature’s shape: a polygon for an area, a line
-for a way that isn’t an area, a point for a node.
+already use. Both columns use [OGC:CRS84](https://en.wikipedia.org/wiki/World_Geodetic_System)
+(WGS84 longitude/latitude, the same coordinates GPS uses) — GeoParquet’s
+default CRS, so it isn’t spelled out explicitly in the file. `atp_geometry`
+is whatever AllThePlaces’ own scrape provides for that feature — nearly
+always a point, though a handful of sources provide lines or polygons
+instead. `osm_geometry` is the matched OpenStreetMap feature’s shape: a
+polygon for an area, a line for a way that isn’t an area, a point for a
+node.
 
-`osm.geometry` isn’t always byte-for-byte what’s in OpenStreetMap,
+`osm_geometry` isn’t always byte-for-byte what’s in OpenStreetMap,
 though: we automatically repair certain geometry errors (like
 self-intersecting lines), and if a shape would otherwise end up with a
 very large number of coordinates, we simplify it — trading a small
 amount of positional accuracy for a smaller file, currently using a
 topology-preserving variant of the
 [Visvalingam–Whyatt algorithm](https://en.wikipedia.org/wiki/Visvalingam%E2%80%93Whyatt_algorithm)
-(this may change in the future). Treat `osm.geometry` as good enough
+(this may change in the future). Treat `osm_geometry` as good enough
 to display — e.g. to mark the affected OSM feature on a map — but not
 as a guaranteed-exact copy of OpenStreetMap’s own data. If you need to
 generate an actual edit against OpenStreetMap, use `way_members`/
@@ -92,7 +110,7 @@ sort key itself isn’t exposed as a column.
 ```sh
 duckdb -c "
 INSTALL spatial; LOAD spatial;
-SELECT osm.type, osm.id, atp.fetched.spider, ST_GeometryType(osm.geometry) AS osm_shape
+SELECT osm.type, osm.id, atp.fetched.spider, ST_GeometryType(osm_geometry) AS osm_shape
 FROM read_parquet('conflated.parquet')
 WHERE osm.id IS NOT NULL
 LIMIT 3;
--- a/docs/outputs/CONFLATED_PARQUET.md
+++ b/docs/outputs/CONFLATED_PARQUET.md
@@ -74,10 +74,11 @@ A few things worth calling out that aren’t obvious from the above:
 
 `atp_geometry`/`osm_geometry` are standard [OGC Simple
 Features](https://postgis.net/workshops/postgis-intro/geometries.html)
-geometries — the same model QGIS, PostGIS, and most other GIS tools
-already use. `atp_geometry` is whatever AllThePlaces’ own scrape
-provides for that feature — nearly always a point, though a handful of
-sources provide lines or polygons instead. `osm_geometry` is the
+geometries — the same model QGIS and PostGIS use, and that formats
+like GeoJSON and GeoPackage are built on too. `atp_geometry` is
+whatever AllThePlaces’ upstream source provides for that feature —
+nearly always a point, though a handful of sources provide lines or
+polygons instead. `osm_geometry` is the
 matched OpenStreetMap feature’s shape: a polygon for an area, a line
 for a way that isn’t an area, a point for a node. Both columns use
 [OGC:CRS84](https://en.wikipedia.org/wiki/World_Geodetic_System)
@@ -88,7 +89,10 @@ default CRS, so it isn’t spelled out explicitly in the file.
 though: we automatically repair certain geometry errors (like
 self-intersecting lines), and if a shape would otherwise end up with a
 very large number of coordinates, we simplify it — trading a small
-amount of positional accuracy for a smaller file, currently using a
+amount of positional accuracy so that you, the downstream client, get
+a geometry that’s directly usable for display without having to
+simplify it yourself first (say, before sending it to a mobile phone
+over a cellular connection). We currently do this using a
 topology-preserving variant of the
 [Visvalingam–Whyatt algorithm](https://en.wikipedia.org/wiki/Visvalingam%E2%80%93Whyatt_algorithm)
 (this may change in the future). Treat `osm_geometry` as good enough
--- a/docs/outputs/CONFLATED_PARQUET.md
+++ b/docs/outputs/CONFLATED_PARQUET.md
@@ -75,14 +75,14 @@ A few things worth calling out that aren’t obvious from the above:
 `atp_geometry`/`osm_geometry` are standard [OGC Simple
 Features](https://postgis.net/workshops/postgis-intro/geometries.html)
 geometries — the same model QGIS, PostGIS, and most other GIS tools
-already use. Both columns use [OGC:CRS84](https://en.wikipedia.org/wiki/World_Geodetic_System)
+already use. `atp_geometry` is whatever AllThePlaces’ own scrape
+provides for that feature — nearly always a point, though a handful of
+sources provide lines or polygons instead. `osm_geometry` is the
+matched OpenStreetMap feature’s shape: a polygon for an area, a line
+for a way that isn’t an area, a point for a node. Both columns use
+[OGC:CRS84](https://en.wikipedia.org/wiki/World_Geodetic_System)
 (WGS84 longitude/latitude, the same coordinates GPS uses) — GeoParquet’s
-default CRS, so it isn’t spelled out explicitly in the file. `atp_geometry`
-is whatever AllThePlaces’ own scrape provides for that feature — nearly
-always a point, though a handful of sources provide lines or polygons
-instead. `osm_geometry` is the matched OpenStreetMap feature’s shape: a
-polygon for an area, a line for a way that isn’t an area, a point for a
-node.
+default CRS, so it isn’t spelled out explicitly in the file.
 
 `osm_geometry` isn’t always byte-for-byte what’s in OpenStreetMap,
 though: we automatically repair certain geometry errors (like
--- a/docs/outputs/CONFLATED_PARQUET.md
+++ b/docs/outputs/CONFLATED_PARQUET.md
@@ -85,7 +85,7 @@ for a way that isn’t an area, a point for a node. Both columns use
 (WGS84 longitude/latitude, the same coordinates GPS uses) — GeoParquet’s
 default CRS, so it isn’t spelled out explicitly in the file.
 
-`osm_geometry` isn’t always byte-for-byte what’s in OpenStreetMap,
+`osm_geometry` isn’t always point-for-point what’s in OpenStreetMap,
 though: we automatically repair certain geometry errors (like
 self-intersecting lines), and if a shape would otherwise end up with a
 very large number of coordinates, we simplify it — trading a small
--- a/src/geometry/geometry_tally.rs
+++ b/src/geometry/geometry_tally.rs
@@ -1,0 +1,66 @@
+//! Bookkeeping for a WKB geometry column written to some output file:
+//! how many of each [`WkbGeometryType`] ended up in it, and the single
+//! largest geometry seen, by byte size.
+
+use super::{WkbGeometryType, wkb_geometry_type};
+
+/// Accumulated while writing one WKB geometry column: a count per
+/// [`WkbGeometryType`] (e.g. becomes a GeoParquet file's `geo` metadata
+/// `geometry_types`, or an INFO log line), and the single largest
+/// geometry seen so far, by byte size, together with a caller-supplied
+/// label identifying which row it came from (e.g. an OSM `type/id`
+/// string, or a spider name) -- so an unusually large geometry can be
+/// tracked down without having to scan the whole file.
+#[derive(Default)]
+pub struct GeometryTally {
+    counts: [u64; WkbGeometryType::ALL.len()],
+    largest_bytes: usize,
+    largest_label: String,
+}
+
+impl GeometryTally {
+    /// Records one geometry. `label` is only called -- so only pays for
+    /// whatever it allocates -- when `wkb` turns out to be the new
+    /// largest seen so far.
+    pub fn record(&mut self, wkb: &[u8], label: impl FnOnce() -> String) {
+        let index = WkbGeometryType::ALL
+            .iter()
+            .position(|t| *t == wkb_geometry_type(wkb))
+            .expect("WkbGeometryType::ALL covers every type wkb_geometry_type can return");
+        self.counts[index] += 1;
+        if wkb.len() > self.largest_bytes {
+            self.largest_bytes = wkb.len();
+            self.largest_label = label();
+        }
+    }
+
+    /// Distinct types actually seen, as GeoParquet `geometry_types`
+    /// strings -- e.g. `["Point"]`, or `["LineString", "Polygon"]`.
+    pub fn geoparquet_types(&self) -> Vec<&'static str> {
+        WkbGeometryType::ALL
+            .iter()
+            .zip(self.counts.iter())
+            .filter(|&(_, &count)| count > 0)
+            .map(|(t, _)| t.geoparquet_name())
+            .collect()
+    }
+
+    /// Logs this tally's per-type counts and largest geometry seen, at
+    /// INFO, as one structured record. `log::info!`'s field list has to
+    /// be literal field names, so the indices below are hardcoded --
+    /// they line up with [`WkbGeometryType::ALL`]'s declaration order.
+    pub fn log(&self, message: &str) {
+        log::info!(
+            point = self.counts[0],
+            line_string = self.counts[1],
+            polygon = self.counts[2],
+            multi_point = self.counts[3],
+            multi_line_string = self.counts[4],
+            multi_polygon = self.counts[5],
+            geometry_collection = self.counts[6],
+            largest_bytes = self.largest_bytes,
+            largest = self.largest_label.as_str();
+            "{}", message
+        );
+    }
+}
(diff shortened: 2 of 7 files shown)


docs/outputs/CONFLATED_PARQUET.md
```diff
@@ -57,23 +72,26 @@ A few things worth calling out that aren’t obvious from the above:
 
 ## Geometry
 
-`atp.geometry`/`osm.geometry` are standard [OGC Simple
+`atp_geometry`/`osm_geometry` are standard [OGC Simple
 Features](https://postgis.net/workshops/postgis-intro/geometries.html)
 geometries — the same model QGIS, PostGIS, and most other GIS tools
-already use. `atp.geometry` is whatever AllThePlaces’ own scrape
+already use. `atp_geometry` is whatever AllThePlaces’ own scrape
```
brawer: "AllThePlaces’ own scrape" -> "AllThePlaces’ upstream source"

docs/outputs/CONFLATED_PARQUET.md
```diff
@@ -57,23 +72,26 @@ A few things worth calling out that aren’t obvious from the above:
 
 ## Geometry
 
-`atp.geometry`/`osm.geometry` are standard [OGC Simple
+`atp_geometry`/`osm_geometry` are standard [OGC Simple
 Features](https://postgis.net/workshops/postgis-intro/geometries.html)
 geometries — the same model QGIS, PostGIS, and most other GIS tools
```
brawer: Also mention GeoJSON and GeoPackage. Actually, these are not tools but formats; find a suitable phrasing.

docs/outputs/CONFLATED_PARQUET.md
```diff
@@ -57,23 +72,26 @@ A few things worth calling out that aren’t obvious from the above:
...
+[OGC:CRS84](https://en.wikipedia.org/wiki/World_Geodetic_System)
+(WGS84 longitude/latitude, the same coordinates GPS uses) — GeoParquet’s
+default CRS, so it isn’t spelled out explicitly in the file.
 
-`osm.geometry` isn’t always byte-for-byte what’s in OpenStreetMap,
+`osm_geometry` isn’t always byte-for-byte what’s in OpenStreetMap,
 though: we automatically repair certain geometry errors (like
 self-intersecting lines), and if a shape would otherwise end up with a
 very large number of coordinates, we simplify it — trading a small
 amount of positional accuracy for a smaller file, currently using a
```
brawer: This is actually not about our own file size, we have enough disk space. We keep an eye on geometry size (and simplify when necessary) is that "you", ie. downstream clients, get a geometry that is directly suitable for display, such as when sending it to a mobile phone over a cellular network. Please rephrase this comment to frame it as user benefit (it really is for the user that we're doing this, not for ourselves). — I know it's not really releated to GeoParquet 2.0, but let's just add this little docfix while editing the file.

src/pipeline/conflate/writer.rs
```diff
@@ -43,6 +45,12 @@ pub struct ParquetWriter {
     atp_fetched_timestamps: TimestampMillisecondBuilder,
     atp_fetched_spiders: StringBuilder,
     atp_geometries: BinaryBuilder,
+    // Distinct WKB geometry types seen on this side so far (e.g.
+    // "Point", "Polygon") -- becomes this column's `geometry_types` in
+    // the GeoParquet `geo` metadata written in `close()`. A `BTreeSet`,
+    // not a `HashSet`, so the metadata's array is written in a
+    // deterministic order across runs.
+    atp_geometry_types: BTreeSet<&'static str>,
```
brawer: Working with strings here is inefficient. Instead, define a custom lightweight enum somewhere in src/geometry, and use that as the key; likewise for the osm_geometry_types counterpart to atp_geometry_types. While at it, make this a counter, and log to INFO how many Points, Multipoints, Lines, MultiLines, ..., GeometryCollections we have written into conflated.parquet as geometries for ATP and OSM, respectively.

src/pipeline/conflate/writer.rs
```diff
@@ -109,6 +119,18 @@ pub struct ParquetRow {
 /// `org.cyclonedx.bom` follows that same reverse-DNS-style namespacing.
 const PROVENANCE_KEY: &str = "org.cyclonedx.bom";
 
+/// Key under which this file's [GeoParquet](https://geoparquet.org/)
+/// metadata is stored -- required by the spec to be named exactly
+/// `geo`. Targets [GeoParquet 2.0-rc.1](https://github.com/opengeospatial/geoparquet/blob/v2.0.0-rc.1/format-specs/geoparquet.md),
+/// the newest available at the time this was written (2.0.0 hasn't
+/// had a final release yet, but the RC's on-disk format is what the
+/// eventual 2.0.0 will also expect -- the `version` field itself is a
+/// fixed `"2.0.0"` string in both). Unlike `PROVENANCE_KEY` above,
```
brawer: Should we also set `org.cyclonedx.bom` _after_ writing the data? As you note, postponement is needed for `geo`, but now we have two different places where metadata is set. It shouldn't make any difference on the output file, but the code would be easier to understand if all kinds of metadata were flowing through the same path.

src/pipeline/conflate/writer.rs
```diff
@@ -367,11 +395,58 @@ impl ParquetWriter {
...
+    /// metadata](https://github.com/opengeospatial/geoparquet/blob/v2.0.0-rc.1/format-specs/geoparquet.md#metadata),
+    /// naming `osm_geometry` as the `primary_column` -- OpenStreetMap is
+    /// the dataset a reader is actually meant to correct, so it's the
+    /// more natural default geometry for a GeoParquet-aware tool to
+    /// display. `crs` is deliberately omitted on both columns: it
+    /// defaults to OGC:CRS84 per spec, which is the same CRS
+    /// `new_geo_field` requests for the columns' native Parquet
+    /// `GEOGRAPHY` logical type -- so the two stay consistent (both
+    /// resolving to OGC:CRS84) without having to spell out a PROJJSON
+    /// object here. (Empirically, `parquet`-rs 59.2 doesn't actually
```
brawer: Is this a bug in the Apache parquet crate? If so, we should file a bug report. With a small, isolated snippet to reproduce the bug, so the Parquet maintainers don't have to understand our pipeline.

src/pipeline/conflate/writer.rs
```diff
@@ -644,6 +729,27 @@ fn wkb(shape: &geo::Geometry<f64>) -> Vec<u8> {
     buf
 }
 
+/// The GeoParquet `geometry_types` name for a WKB-encoded geometry
+/// (e.g. `"Point"`, `"Polygon"`, `"MultiPolygon"`) -- read straight off
+/// the WKB header (byte order + type code) via `wkb::reader`, without
+/// decoding any coordinates.
+fn wkb_geometry_type_name(wkb: &[u8]) -> Result<&'static str> {
+    use ::wkb::reader::GeometryType;
```
brawer: Reading the WKB string into a geo::Geometry is not cheap. Instead, let's take the geometry code from the `wkb` slice. Something like https://github.com/georust/wkb/blob/a495fb4b0ce553eb4c26f1376ca701c6968f5d89/src/common.rs#L99-L143 but simpler (we just need to know the geometry type, we don't care about SRS, M and Z coordinates, etc.). If I remember correctly, wkb[0] is just endianness; since we produce the WKB slice ourselves, we already know it's little-endian (even if we ever run on a big-endian machine), so we can just panic! if that's not the case, no need to implement a big-endian decoder.

src/pipeline/conflate/writer.rs
```diff
@@ -54,6 +62,8 @@ pub struct ParquetWriter {
     osm_way_members: ListBuilder<UInt64Builder>,
     osm_relation_members: ListBuilder<StructBuilder>,
     osm_geometries: BinaryBuilder,
+    // See `atp_geometry_types` above.
+    osm_geometry_types: BTreeSet<&'static str>,
```
brawer: Likewise.

src/pipeline/conflate/writer.rs
brawer: Make this a counter instead of a set. At the end, log to INFO how many Points, Multipoints, Lines, MultiLines, ..., GeometryCollections we have written into conflated.parquet. Write separate log entries for atp_geometry and osm_geometry. For OSM, also log what OSM feature (as type/id string, like "relation/123") had the largest number of bytes in osm_geometry, so we can track it down if we ever run into problems with excessive geometries. Likewise for ATP (ATP doesn't really have stable feature IDs, but at least we can log the spider, this will provide a good enough enough clue should we ever have geometry size problems on the ATP side).

docs/outputs/CONFLATED_PARQUET.md
brawer: Done in 9b8feb5.

docs/outputs/CONFLATED_PARQUET.md
brawer: Done in 9b8feb5: "the same model QGIS and PostGIS use, and that formats like GeoJSON and GeoPackage are built on too."

docs/outputs/CONFLATED_PARQUET.md
brawer: Done in 9b8feb5 -- reframed around the downstream-client benefit (display-ready geometry, e.g. before sending to a phone over cellular), kept short.

src/pipeline/conflate/writer.rs
brawer: Agreed, and you're right that there's no ready-made enum for just this in geo/geo_types/wkb -- added `WkbGeometryType` in new src/geometry/wkb_type.rs (7 variants, Debug/Display via geoparquet_name(), an ALL const for building zero-initialized tallies). `wkb_geometry_type()` reads the type straight off the WKB header (byte order + type code), no coordinate decode -- this also addresses your other comment below about not going through geo::Geometry. Done in 9b8feb5.

src/pipeline/conflate/writer.rs
brawer: Agreed, done in 9b8feb5: provenance_bom is now just stored on ParquetWriter in create(), and both PROVENANCE_KEY and GEO_METADATA_KEY get set via append_key_value_metadata in close(), right next to each other.

src/pipeline/conflate/writer.rs
brawer: Investigated with an isolated repro (plain arrow/parquet/parquet-geospatial, no dependency on our pipeline) -- it's not a bug. parquet's own `logical_type_for_binary` (parquet 59.2, src/arrow/schema/extension.rs) deliberately canonicalizes a lon/lat CRS and `Spherical` edges to `None` when building the native LogicalType, because those are themselves the Parquet spec's own defaults for GEOGRAPHY -- the same omit-if-default encoding GeoParquet's own `geo` metadata uses for `crs`. So `crs: None, algorithm: None` in the native logical type is the correct, intentional encoding of exactly the OGC:CRS84/spherical values we requested, not data loss. No bug report needed. Corrected the doc comment (which previously called this an unconfirmed gap) in 9b8feb5.

src/pipeline/conflate/writer.rs
brawer: Done in 9b8feb5 -- see reply above on the enum comment; `wkb_geometry_type()` in the new src/geometry/wkb_type.rs reads byte 0 (asserts little-endian, panics otherwise per your suggestion) and the type code at bytes 1..5 directly, no geo::Geometry decode.

src/pipeline/conflate/writer.rs
brawer: Same fix, done in 9b8feb5 (see reply on the sibling atp_geometry_types comment).

src/pipeline/conflate/writer.rs
brawer: Done in 9b8feb5: GeometryTally now holds a [u64; 7] counter plus largest-seen (bytes + a lazily-computed label -- OSM type/id like "way/608979139", ATP spider -- only formatted when a row actually becomes the new largest, not per row). Logged at INFO in close(), one record each for atp_geometry/osm_geometry: `conflate.write: atp_geometry geometry types` / `conflate.write: osm_geometry geometry types`, with point/line_string/polygon/multi_point/multi_line_string/multi_polygon/geometry_collection/largest_bytes/largest fields.

docs/outputs/CONFLATED_PARQUET.md
```diff
@@ -57,23 +72,30 @@ A few things worth calling out that aren’t obvious from the above:
...
+polygons instead. `osm_geometry` is the
 matched OpenStreetMap feature’s shape: a polygon for an area, a line
-for a way that isn’t an area, a point for a node.
+for a way that isn’t an area, a point for a node. Both columns use
+[OGC:CRS84](https://en.wikipedia.org/wiki/World_Geodetic_System)
+(WGS84 longitude/latitude, the same coordinates GPS uses) — GeoParquet’s
+default CRS, so it isn’t spelled out explicitly in the file.
 
-`osm.geometry` isn’t always byte-for-byte what’s in OpenStreetMap,
+`osm_geometry` isn’t always byte-for-byte what’s in OpenStreetMap,
```
brawer: "byte-for-byte" -> "point-for-point" (it's a bit misleading to talk about bytes here)

src/geometry/wkb_type.rs
```diff
@@ -0,0 +1,144 @@
...
+        Self::LineString,
+        Self::Polygon,
+        Self::MultiPoint,
+        Self::MultiLineString,
+        Self::MultiPolygon,
+        Self::GeometryCollection,
+    ];
+
+    /// The GeoParquet `geometry_types` string for this type, e.g.
+    /// `"Point"`, `"MultiPolygon"`.
```
brawer: Add comment: See "geometry_types" in https://geoparquet.org/releases/v2.0.0-rc.1/

src/pipeline/conflate/writer.rs
```diff
@@ -109,10 +125,84 @@ pub struct ParquetRow {
...
+/// the newest available at the time this was written (2.0.0 hasn't
+/// had a final release yet, but the RC's on-disk format is what the
+/// eventual 2.0.0 will also expect -- the `version` field itself is a
+/// fixed `"2.0.0"` string in both). Unlike `PROVENANCE_KEY` above,
+/// this can't be built until every row has been written -- see
+/// `close()` -- because `geometry_types` below depends on what
+/// geometries actually ended up in the file.
+const GEO_METADATA_KEY: &str = "geo";
+
+/// Per-column bookkeeping accumulated while writing one of the two
```
brawer: Ooh, nice! Please move GeometryTally into the crate-level geometry module, rephrasing the doc to be more general (not talking about conflated.parquet with its two columns). Then, file a low-priority GitHub issue that the ATP phase of the pipeline should do the same tallying (as in conflation) when it writes alltheplaces.parquet, producing a similar log record, using this new functionality in mod geometry.

docs/outputs/CONFLATED_PARQUET.md
brawer: Done in 0ef0a8c.

src/geometry/wkb_type.rs
brawer: Done in 0ef0a8c -- linked, with a note that the page has no anchors yet.

src/pipeline/conflate/writer.rs
brawer: Done in 0ef0a8c: GeometryTally moved to src/geometry/geometry_tally.rs, re-exported as crate::geometry::GeometryTally, doc comment generalized away from conflated.parquet's two columns.

Filed #690 for the alltheplaces.parquet side -- but it turned out to be more than just missing tallying. While tracing where atp_geometry's WKB actually comes from, found that atp::mod::find_point() collapses every ATP feature (Point, LineString, Polygon, whatever the source gives) down to a single representative point before it ever reaches Place, and alltheplaces.parquet has no geometry column at all today -- so alltheplaces.parquet's own geometry tally would, right now, always just report 100% Point, which wouldn't be worth the code. #690 covers giving Place a real geometry field first (so atp_geometry in conflated.parquet actually reflects the source data, per its own doc comment's claim), with the GeometryTally/logging parity bundled into the same piece of work once that lands. @brawer: next thing after this PR, before cutting the release, per your call.

src/geometry/geometry_tally.rs
```diff
@@ -0,0 +1,66 @@
...
+//! largest geometry seen, by byte size.
+
+use super::{WkbGeometryType, wkb_geometry_type};
+
+/// Accumulated while writing one WKB geometry column: a count per
+/// [`WkbGeometryType`] (e.g. becomes a GeoParquet file's `geo` metadata
+/// `geometry_types`, or an INFO log line), and the single largest
+/// geometry seen so far, by byte size, together with a caller-supplied
+/// label identifying which row it came from (e.g. an OSM `type/id`
+/// string, or a spider name) -- so an unusually large geometry can be
```
brawer: "spider name" -> "AllThePlaces spider name"

src/geometry/geometry_tally.rs
```diff
@@ -0,0 +1,66 @@
...
+    counts: [u64; WkbGeometryType::ALL.len()],
+    largest_bytes: usize,
+    largest_label: String,
+}
+
+impl GeometryTally {
+    /// Records one geometry. `label` is only called -- so only pays for
+    /// whatever it allocates -- when `wkb` turns out to be the new
+    /// largest seen so far.
+    pub fn record(&mut self, wkb: &[u8], label: impl FnOnce() -> String) {
```
brawer: Explain what `wkb` means: Well-Known Binary (WKB). This may not be obvious to people who aren't familiar with geospatial computing. Admitteldy, someone unfamiliar with geospatial software is unlikely to ever look a this code; but why not add a super-brief comment for whose where `wkb` doesn't ring a bell.

src/geometry/geometry_tally.rs
brawer: Done in ea8e756.

src/geometry/geometry_tally.rs
brawer: Done in ea8e756 -- explained in the module doc comment, linked to libgeos.org's WKB spec page.

Issue #690, opened 2026-08-18
ATP phase discards non-point geometry -- Place/alltheplaces.parquet/conflated.parquet's atp_geometry should carry the source's real shape


brawer:
### Summary

`atp_geometry` in `conflated.parquet` is documented (`docs/outputs/CONFLATED_PARQUET.md`) as "whatever AllThePlaces' upstream source provides for that feature — nearly always a point, though a handful of sources provide lines or polygons instead." That's aspirational, not actual: today, every ATP feature is reduced to a single representative point before it ever reaches `Place`, so `atp_geometry` is *always* a point, regardless of what AllThePlaces' own scrape actually provided.

Traced during review of #688 (empirically, not by inspection alone: checked every call site of `Place::shape()` — only `pipeline::conflate::writer::ParquetRow::from_conflated` consumes it for real, so the blast radius of fixing this is small).

### Root cause

- `atp::mod::find_point()` (`src/atp/mod.rs`) takes whatever `geo::Geometry` a GeoJSON feature parses to (`Point`, `LineString`, `MultiLineString`, `Polygon`, ...) and immediately collapses it to a single `geo::Point` — the line's midpoint (`Haversine::point_at_ratio_from_start(..., 0.5)`) or a polygon's `interior_point()` — before `make_place()` ever constructs a `Place`.
- `Place` (`src/places/mod.rs`) only stores `s2_cell_id` (an S2 cell derived from that single point) — there's no field for real geometry at all.
- `Place::shape()` doesn't return stored geometry; it *reconstructs* a `geo::Point` from `s2_cell_id`'s centroid. So even the reduction in `find_point` is compounded — `Place` doesn't preserve the exact reduced point either, only the S2 cell it falls in.
- `alltheplaces.parquet` (`src/places/writer.rs`) doesn't have a geometry column at all today — just `s2_cell_id`, `mask`, `spider`, `tags`, `fetched`.

### What to do

[...]

PR #691, opened 2026-08-18
Move src/atp into src/pipeline/atp

brawer:
The AllThePlaces import is just another pipeline stage, same as `src/pipeline/osm/` and `src/pipeline/conflate/` -- there's no reason for it to live at the crate root instead of alongside its siblings.

Pure move (`git mv`, no content changes to the `atp` files themselves) plus wiring updates:

- `src/lib.rs`: drop the top-level `mod atp;`.
- `src/pipeline/mod.rs`: add `mod atp;`, alongside `conflate`/`edits`/`geostats`/`osm`. Re-export `AtpMetadata` and `read_cached_metadata` (renamed to `read_cached_atp_metadata`, since `osm` already has its own `read_cached_metadata` re-exported here -- same name, different metadata type) for `provenance.rs` to use, matching the existing pattern for `osm`'s own re-exports. Internal call sites (`crate::atp::import_atp`, `crate::atp::collect_wikidata_ids`) become plain `atp::...` now that they're sibling modules.
- `src/provenance.rs`: `crate::atp::read_cached_metadata` -> `crate::pipeline::read_cached_atp_metadata`.
- `tests/integration_test.rs`: one stale path in a comment.

No behavior change. Requested ahead of #690 (which will touch this code) to land as a clean, isolated diff first.

## Verification

`cargo build`, `cargo test`, `cargo clippy --all-targets`, `cargo fmt --check` all clean.


brawer (commented):
LGTM

PR #692, opened 2026-08-18
ATP phase: preserve real (non-point) geometry

### LINKED ISSUE #690: ATP phase discards non-point geometry -- Place/alltheplaces.parquet/conflated.parquet's atp_geometry should carry the source's real shape

brawer:
Fixes #690.

## Root cause

`atp::mod::find_point()` reduced every feature's geometry — point, line, polygon, whatever the source gives — to a single representative point before it ever reached `Place`, and `Place`/`alltheplaces.parquet` had no geometry field at all. So `conflated.parquet`'s `atp_geometry` was always a point, contradicting its own doc comment in `CONFLATED_PARQUET.md`.

## What changed

- `Place` gets a real `shape_wkb: Vec<u8>` field (WKB, not `geo::Geometry` directly — `f64` doesn't implement `Eq`/`Ord`, which `Place` needs for its external sort). `s2_cell_id` stays exactly as before: a spatial-index sort/query key derived from one representative point — a separate concern from the feature's actual shape now that the two are no longer conflated. `Place::shape()` decodes `shape_wkb` instead of reconstructing a point from `s2_cell_id`.
- `pipeline::atp::make_place` keeps the parsed `geo::Geometry` instead of discarding it after `find_point()` reduces it to a point; `find_point` itself now takes `&geo::Geometry`, not a whole `GeoJson` document (a new `parse_geometry` does that parse once, shared by both).
- `alltheplaces.parquet` gets a `"shape"` column (plain WKB bytes, same internal/ephemeral-format reasoning as the existing `"fetched"` column — no GeoParquet-style logical type needed for a file nothing outside this pipeline reads).
- `places::writer::ParquetWriter` now tallies geometry types the same way `pipeline::conflate::writer` does (#688), logging at INFO in `close()` — using the `crate::geometry::GeometryTally`/`WkbGeometryType` machinery that was deliberately moved to crate level in #688 for exactly this reuse.
- New `crate::geometry::write_wkb`, deduped out of `pipeline::conflate::writer`'s own local `wkb()` (now shared with `Place::new` and wherever else geometry needs encoding).
- Regenerated `tests/test_data/alltheplaces.parquet` — required for the new `"shape"` column, but this also fixed a real, pre-existing staleness bug found along the way: the checked-in fixture had drifted out of sync with `alltheplaces.zip` a while ago (7 rows, `"atp/"`-prefixed spider names, one uniform `fetched` timestamp — none of which match the current zip), undetected because `places::reader`'s tests only checked loose invariants (non-zero cell id, non-empty spider, well-formed tags). Regenerated by actually running `import_atp` against the real zip fixture, and tightened the two affected tests' assertions to match (including `pipeline::atp::wikidata_ids`'s, which hardcoded the old fixture's Wikidata-ID set).

`docs/outputs/CONFLATED_PARQUET.md`'s existing "nearly always a point, though a handful of sources provide lines or polygons instead" claim needed no edit — it's now actually true instead of aspirational, per prior discussion on #690.

## Verification

- `cargo build`/`test`/`clippy --all-targets`/`fmt --check` all clean (246 lib tests + 4 integration tests).
- Ran the real pipeline against the test fixtures and checked: `gpio check spec`/`check all` on the resulting `conflated.parquet` still **49 passed, 0 warnings, 0 failed**; the new `"import_atp: alltheplaces.parquet geometry types"` log line shows up correctly in `pipeline.log`.
- New regression test `pipeline::atp::tests::test_make_place_preserves_non_point_geometry` — asserts a LineString shape survives `make_place()`, which the old find-and-discard behavior could not produce (verified this actually fails against the pre-fix code, not just that it passes now).
- Today's real `alltheplaces.zip` fixture only has point geometry (matches @brawer's note that non-point ATP data is currently rare — `bern_ch` being about the only real-world example), so this PR's own test fixtures don't exercise non-point geometry end-to-end through the full pipeline; the regression test above covers that path directly at the `make_place` level instead.


src/geometry/wkb_writer.rs
```diff
@@ -0,0 +1,32 @@
+//! Encodes a `geo::Geometry` as WKB (Well-Known Binary -- see
+//! `geometry_tally`'s module comment).
```
brawer: Move the mentioned comment to here, and change geometry_tally's comment to point to wkb_writer here.

src/geometry/wkb_writer.rs
```diff
@@ -0,0 +1,32 @@
...
+
+use geo::Geometry;
+
+/// Encodes `shape` as little-endian WKB.
```
brawer: Explain why we have our own wrapper around wkb::writer: Optimized for points (thus, fewer memory allocations), always little-endian (thus, simpler API).

src/places/reader.rs
```diff
@@ -151,9 +163,16 @@ mod tests {
...
 
-    /// Test file with 7 places -- see the old `PlaceIndex` tests (git
-    /// history) for the full per-place s2_cell_id/mask breakdown; this
-    /// only needs the total count and that every place decodes cleanly.
+    /// Test file with 10 places -- produced by running the real
+    /// `import_atp` pipeline stage against `tests/test_data/
+    /// alltheplaces.zip` (see `tests/integration_test.rs`'s own use of
+    /// that same zip), not hand-crafted; this only needs the total count
+    /// and that every place decodes cleanly. (A previous version of this
+    /// fixture had drifted out of sync with the zip -- 7 places,
```
brawer: No need for historical details, remove the part "-- 7 places, [...] timestamp --".

src/geometry/wkb_writer.rs
brawer: Done in 68d1738.

src/geometry/wkb_writer.rs
brawer: Done in 68d1738.

src/places/reader.rs
brawer: Done in 68d1738.

brawer (commented):
LGTM

PR #693, opened 2026-08-18
GeometryTally::log: take target as an explicit parameter

brawer:
Fixes a logging regression found during a post-release-prep audit (not by review comment).

## The bug

`GeometryTally::log()` called `log::info!()` without an explicit `target`, so it defaulted to wherever the macro is physically written — always `geometry::geometry_tally`, regardless of which caller (`pipeline::conflate::writer` vs. `places::writer`) is actually doing the logging. This came in as a side effect of moving `GeometryTally` from `pipeline::conflate::writer` into the shared `crate::geometry` module in #690/#692 — that move itself was correct, this is a follow-up fixing what came with it.

`docs/LOGGING.md` documents `target` as "which module logged it"; a shared helper silently defeated that for both of its current callers, collapsing two genuinely different log sources onto the same `target` value (the `message` text still differed, so this wasn't catastrophic, just a real loss of fidelity for anyone filtering/grepping logs by module).

## Fix

`GeometryTally::log(&self, target: &str, message: &str)` now takes `target` explicitly; both call sites pass `module_path!()`, exactly reproducing what `log::info!` would have resolved to if it were written directly at each call site (as it was, before the move).

## Verification

- `cargo build`/`test`/`clippy --all-targets`/`fmt --check` all clean.
- Empirically: ran the pipeline against the test fixtures and checked `pipeline.log` directly. Before this fix, both `conflate.write: ...` and `import_atp: ...` geometry-type log lines showed `"target":"osm_diffs::geometry::geometry_tally"`. After: `osm_diffs::pipeline::conflate::writer` and `osm_diffs::places::writer` respectively, matching their real callers.

### DIFF
--- a/src/geometry/geometry_tally.rs
+++ b/src/geometry/geometry_tally.rs
@@ -51,8 +51,18 @@ impl GeometryTally {
     /// INFO, as one structured record. `log::info!`'s field list has to
     /// be literal field names, so the indices below are hardcoded --
     /// they line up with [`WkbGeometryType::ALL`]'s declaration order.
-    pub fn log(&self, message: &str) {
+    ///
+    /// `target` is taken as an explicit parameter -- callers should pass
+    /// `module_path!()` -- rather than left to `log::info!`'s own
+    /// default, which would resolve to wherever this macro call is
+    /// physically written (i.e. always `geometry::geometry_tally`,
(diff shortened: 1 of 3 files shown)


brawer (commented):
LGTM

PR #694, opened 2026-08-18
Consolidate the osm_id*10+{type} encode/decode into shared helpers

brawer:
Found during a post-release-prep audit, filed as no issue existed for it (a code comment on \`decode_member_id\` cited #662, but #662 turned out to be about something unrelated — a centroid pre-check idea).

## The problem

\`Feature.id\`'s encoding (\`osm_id * 10 + {1=node, 2=way, 3=relation}\`) was independently reimplemented in at least six places — \`assemble.rs\`'s \`make_feature_id\`/\`feature_to_osm_id\`/\`debug_id_str\`, \`prune.rs\`'s several inline \`id * 10 + N\`/\`id % 10\` sites, and \`conflate/writer.rs\`'s inline match plus its own \`decode_member_id\`. Same encoding, same magic numbers, six chances to get it wrong.

## What changed

New shared helpers in \`pipeline::osm\` (\`encode_feature_id\`, \`decode_feature_id\`, \`osm_type_str\`), re-exported crate-wide the same way \`OsmMetadata\`/\`read_cached_metadata\` already are, for \`conflate::writer\` to use. \`decode_feature_id\` returns \`Option\` rather than panicking, so callers choose their own panic message (or, for \`assemble.rs\`'s \`debug_id_str\`, a safe non-panicking fallback — it exists specifically to build the message for a *different* panic, so it must never itself panic and obscure that).

\`assemble.rs\`/\`prune.rs\`/\`conflate/writer.rs\` all updated to call the shared helpers instead of reimplementing the encoding.

No behavior change — verified empirically, not just by inspection: ran the pipeline against the test fixtures before and after, and \`conflated.parquet\`'s OSM rows (type, id, way/relation member counts) came out byte-identical.

## Verification

- \`cargo build\`/\`test\`/\`clippy --all-targets\`/\`fmt --check\` all clean (249 lib tests, including 3 new ones for the shared helpers themselves, + 4 integration tests — all pre-existing pinned assertions on real fixture data still pass unchanged).
- Ran the real pipeline against the test fixtures and checked \`conflated.parquet\`'s OSM rows and \`gpio check spec\` (still 49 passed/0 warnings/0 failed).

Independent of #693 (the logging-target fix) — separate branches off \`main\`, no shared history.

### DIFF
--- a/src/pipeline/conflate/writer.rs
+++ b/src/pipeline/conflate/writer.rs
@@ -27,8 +27,13 @@ use std::{
 
 use super::ConflatedFeature;
 use crate::{
-    geometry::GeometryTally, matchers::OsmCandidate, tables::StringPool, utils::UtcTimestamp,
+    geometry::GeometryTally,
+    matchers::OsmCandidate,
+    pipeline::{decode_feature_id, osm_type_str},
+    tables::StringPool,
+    utils::UtcTimestamp,
 };
+use osm_pbf_iter::RelationMemberType;
 
 pub struct ParquetWriter {
     path: PathBuf,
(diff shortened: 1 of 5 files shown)


brawer (commented):
LGTM

Issue #696, opened 2026-08-18
Decide which OSM edit-review tool(s) our output should target, and how -- not just "MapRoulette"


brawer:
This repo's README currently states the plan as "automatically feed edit proposals to MapRoulette." That's not actually settled — this issue is to think it through properly and produce a real plan, before building against any one tool as if it were the only option.

## MapRoulette specifically

Already discussed with the MapRoulette team directly; a few things are already clear, the rest isn't:

- **Hosting**: we host the (potentially large) generated challenge GeoJSON ourselves, in MapRoulette's own challenge format, on a publicly accessible URL — likely S3, since this pipeline already uploads its other outputs there and S3 makes them public automatically. Our pipeline then sends a small *authenticated* REST call to MapRoulette itself, telling it to (re-)fetch from that URL. Planned cadence: once a week, so challenges don't go stale.
- **Scale is a real problem, not a hypothetical one**: our data is potentially millions of tasks, far above MapRoulette's normal volume. We need to figure out the actual mechanics — sharding, pacing, whatever it takes — so we don't take MapRoulette down. No plan for this yet.
- **Task quality / reviewer trust**: MapRoulette's own stated concern is that users blindly click "Accept" on tasks without doing any real verification. If that happens with our tasks at scale, bad edits land in OSM. We need a real answer here, not just "we'll write good instructions."

## Splitting/sharding tasks so they're actually reviewable

Open questions, not yet decided:

- Split by store brand?
- Split by which OSM tag is being changed — separate tasks for `opening_hours` vs. `website` vs. `phone` vs. others?
- A separate task category for **new stores**: appears in a brand's store finder (i.e., in the AllThePlaces scrape) but has no matching OSM feature yet.
- A separate task category for **likely-closed stores**: we have a matching OSM feature tagged as this brand, but it's no longer in the brand's store finder (either genuinely closed, or it was never accurately listed there in the first place).

## The bigger question: not every edit belongs on MapRoulette

Different kinds of edits likely suit different review tools, each with a different audience and review model:

- **On-the-ground / field mapping**: [StreetComplete](https://streetcomplete.app/), [Every Door](https://every-door.app/) — for someone physically at the location.
- **Armchair review**: [MapRoulette](https://maproulette.org/), [Osmose](https://wiki.openstreetmap.org/wiki/Osmose) — for someone checking imagery/data remotely.

We need an actual plan for which categories of edits go to which system(s), not an assumption that everything funnels into one tool.

[...]

PR #697, opened 2026-08-18
README: don't commit to MapRoulette specifically, link to #696

brawer:
Which edit-review tool(s) our output should actually target — and how to split/shard tasks so they're reviewable, and whether different kinds of edits belong on different tools (MapRoulette/Osmose vs. StreetComplete/Every Door) — isn't decided yet, per #696. The README shouldn't imply it's settled by naming MapRoulette specifically.


brawer (commented):
LGTM

PR #698, opened 2026-08-18
Document why places and tables are separate, fix stale PlaceIndex reference

brawer:
Prompted by discussing whether \`places\` should move into \`tables\` while reviewing module structure for the design doc: they shouldn't (\`tables\`' own module doc commits to a specific mmap + page-cache design that \`places\` genuinely doesn't need, given ATP's much smaller data volume), but the split deserved a one-line explanation in both modules rather than just being left implicit.

Also fixes \`tables::feature_index\`'s doc comment, which still referenced \`crate::places::PlaceIndex\` — a type retired along with the old OSM-via-\`Place\` path (\`places::reader\` now has the deliberately cache-free \`PlaceReader\` instead). Rewrote the comparison to state the real distinction (\`Place\` is always a single point, so it never needs the coverage-vs-centroid split \`OsmFeatureIndex\` has to handle) instead of naming a type that no longer exists.

### DIFF
--- a/src/places/mod.rs
+++ b/src/places/mod.rs
@@ -37,6 +37,20 @@ pub use writer::ParquetWriter;
 /// An AllThePlaces feature.
 #[derive(Debug, DeepSizeOf, Eq, Ord, PartialEq, PartialOrd, Serialize, Deserialize)]
 pub struct Place {
+    /// A single S2 cell, derived from one representative point --
+    /// `pipeline::atp::find_point`'s reduction of this feature's actual
+    /// geometry (see `shape_wkb` below), regardless of whether that
+    /// geometry is itself a point. Unlike `tables::OsmFeatureIndex`
+    /// (which indexes real OSM geometry by its full multi-cell
+    /// `coverage_s2_cell_id`), `Place` has no equivalent multi-cell
+    /// coverage of its own -- not because it's assumed to be small
+    /// enough to ignore, but because the `s2` crate this pipeline uses
+    /// doesn't yet support computing S2 coverage for lines/polygons,
+    /// only points. `pipeline::conflate` works around that by searching
+    /// a radius around this cell instead of querying by actual coverage
+    /// -- see
+    /// [alltheplaces/osm-diffs#700](https://github.com/alltheplaces/osm-diffs/issues/700)
+    /// for the plan to fix this once the library gains that support.
     pub s2_cell_id: u64,
     pub spider: String,
     pub mask: MatchMask,
@@ -48,10 +62,9 @@ pub struct Place {
     /// `geo::Geometry`, so `Place` can keep deriving `Eq`/`Ord` (needed
     /// for the external sort in `pipeline::atp::process_places`) --
     /// `f64` coordinates don't implement those. Distinct from
-    /// `s2_cell_id`, which is always derived from a single representative
-    /// point regardless of this field's actual geometry type -- that's a
-    /// spatial-index sort/query key, not the feature's real shape. Use
-    /// `shape()` for the decoded `geo::Geometry`.
+    /// `s2_cell_id` (see above), which is always a single cell derived
+    /// from one representative point regardless of this field's actual
+    /// geometry type. Use `shape()` for the decoded `geo::Geometry`.
     pub shape_wkb: Vec<u8>,
 
     /// When AllThePlaces fetched this feature from its upstream source
--- a/src/places/mod.rs
+++ b/src/places/mod.rs
@@ -1,3 +1,14 @@
+//! `Place`, AllThePlaces' own durable on-disk representation --
+//! `alltheplaces.parquet` -- written by `pipeline::atp`, read back by
+//! `pipeline::conflate`/`pipeline::atp::wikidata_ids`, and consumed
+//! directly by `matchers` when scoring a candidate match.
+//!
+//! Plays the same cross-pipeline-stage-storage role `crate::tables`
+//! plays for OSM data, but isn't one of `tables`' mmap'd structures:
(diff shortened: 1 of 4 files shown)


src/places/mod.rs
```diff
@@ -1,3 +1,14 @@
+//! `Place`, AllThePlaces' own durable on-disk representation --
```
brawer: As worded now, this is misleading: The durable represenation of the AllThePlaces weekly runs is a ZIP file, containing the result of each spider as a GeoJSON FeatureCollection, with each contained feature on a separate line. (The first line is the GeoJSON FeatureCollection, including metadata such as collection timestamp, and license). When the osm-diffs pipeline starts up, one of the first things it does is to fetch the last run's ZIP file, parse it, filter anything we don't like (in particular, we ignore any data from sources whose license is not suitable for conflation with OpenStreetMap, according to advice from the OSM Licensing Working Group — see the link in conflated.parquet, embedded as "evidence" into the output BOM), and convert it to a spatially sorted Parquet file. The resulting Parquet file, called `alltheplaces.parquet` is only stored in the working directory of our local disk; we don't upload it anywhere. So it's not a "durable on-disk representation". — Thanks for summarizing my burb into something coherent :-) — please change this comment accordingly.

src/tables/feature_index.rs
```diff
@@ -1,11 +1,12 @@
 //! Spatial index over OSM [Feature] protos, keyed by S2 cell coverage.
 //!
-//! Unlike [crate::places::PlaceIndex] (which indexes single-point `Place`
-//! values backed by Parquet), `OsmFeatureIndex` indexes real OSM
-//! geometry -- points, lines, and polygons alike -- so a single feature
-//! can cover more than one S2 cell. That means the field used to sort
-//! features physically on disk (`centroid_s2_cell_id`, for locality only)
-//! is *not* a valid query key: a query has to test a feature's full
+//! Unlike `Place` (see `crate::places`) -- always a single point, so its
```
brawer: Actually, place can have non-point geometry, it's just rare. At the moment, we take the Place's centroid and search within a cap (up to 2 km, if I remember correctly) to see if this intersects any OSM feature in FeatureIndex. So the assumption isn't really that Place is always a point; only that the Place is small enough so we can just reduce it to a point for the purpose of querying. As of August 2026, the Rust port of the s2 geometry library does not yet support computing the s2 cell coverage for lines and polygons — that's the real reason why we do this. Maybe we should explain this somewhere (here? elsewhere?), and file a bug to watch s2 librar, and change our implementation once s2 has polyline/polygon support. (Funnily, I — ie., @brawer — am currently the only active maintainer of the Rust port of the s2 library, but I don't want to spend much time on it, and I haven't found anyone keen on replacing me, so it's unlikely this will get implemented anytime soon). In any case, the actual impact of this change on osm-diffs will be minimal, because there's only few data sources in AllThePlaces with non-point geometry. — Please summarize, file a low-priorty bug on osm-diffs to watch rust-s2, to be looked at in August 2027, and link to this bug wherever appropriate. Mabe around here, or in pipeline/atp where we compute the s2_cell_id, or wherever else in this codebase you find appropriate.

src/places/mod.rs
brawer: Thanks for the correction -- done in facbc8b. Rewrote to describe the actual ZIP → GeoJSON (per-spider, one feature per line, first line = FeatureCollection metadata) → license-filtered → spatially-sorted Parquet flow, and made clear alltheplaces.parquet itself is local-only, never uploaded, not durable.

src/tables/feature_index.rs
brawer: Fixed the claim, and filed #700 to track watching the s2 crate for line/polygon coverage support (checked the real search-radius constants while I was at it: 10m/400m/1000m by feature category, in matchers::match_distance -- not quite the 2km you recalled, but same idea). Linked #700 from here and from pipeline::atp::find_point, which is where the actual point-reduction happens. Done in facbc8b.

src/tables/feature_index.rs
```diff
@@ -1,13 +1,22 @@
...
-//! really two on-disk structures:
+//! `OsmFeatureIndex` indexes real OSM geometry -- points, lines, and
+//! polygons alike -- so a single feature can cover more than one S2
+//! cell. That means the field used to sort features physically on disk
+//! (`centroid_s2_cell_id`, for locality only) is *not* a valid query
+//! key: a query has to test a feature's full `coverage_s2_cell_id`
+//! list, not just its centroid. So this index is really two on-disk
+//! structures:
+//!
+//! `Place` (see `crate::places`) has no equivalent multi-cell coverage,
```
brawer: This is now a well-written paragraph, but I think you inserted it at the wrong place. At the moment, it interrupts the flow from "this index is really two on-disk structures:", and the two structures. Perhaps move it to the end of the module comment, possibly as a separate section to discuss FeatureIndex vs Places?

brawer: Good catch, done in 3e455f7 -- moved it to its own '# Place has no equivalent' section after the two-structures list and the mmap/read-only paragraph.

src/tables/mod.rs
```diff
@@ -4,6 +4,13 @@
 //! as there is enough disk space, a table can be larger than the
 //! physical RAM installed on the machine. This makes it possible to
 //! process the entire OpenStreetMap planet on cheap worker machine.
+//!
+//! `crate::places` plays the same role for AllThePlaces data -- durable,
```
brawer: Remove "durable" (it's just internal to the pipeline's working dir, never uploaded to durable storage).

brawer: Done in 1007f65.

brawer (commented):
LGTM

PR #699, opened 2026-08-18
Move geostats.rs into the geometry module

brawer:
It's a helper that sets up how \`conflated.parquet\`'s GEOMETRY/GEOGRAPHY columns get their Parquet-native geospatial statistics computed — not a pipeline stage like assemble/prune/tiles/upload, which is what \`pipeline/\` is otherwise for. Already flagged with a \`// TODO: Move into crate::geometry?\` comment; this just does it.

Pure move: \`init()\` re-exported as \`crate::geometry::init_geospatial_stats\` (renamed from the bare \`init\` to be self-explanatory at the call site, now that it's not obviously pipeline-startup-related from its module path alone), \`pipeline::run_pipeline_steps\` updated to call it. Also added a doc comment to \`init()\` itself, since the reasoning for calling it once at startup was previously obvious from being colocated with \`run_pipeline_steps\` and isn't anymore.

Verified: \`cargo build\`/\`test\`/\`clippy --all-targets\`/\`fmt --check\` all clean.

### DIFF
--- a/src/geometry/geostats.rs
+++ b/src/geometry/geostats.rs
@@ -42,6 +42,12 @@ impl GeoStatsAccumulatorFactory for CustomGeoStatsAccumulatorFactory {
     }
 }
 
+/// Installs [`CustomGeoStatsAccumulatorFactory`] as the process-wide
+/// `GeoStatsAccumulatorFactory` (see the module doc comment above for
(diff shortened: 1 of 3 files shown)


brawer (commented):
LGTM

PR #702, opened 2026-08-18
Move provenance.rs into pipeline

brawer:
It's entirely a pipeline detail: its only real caller is `pipeline::conflate` (building `conflated.parquet`'s embedded BOM), and it only ever reads pipeline-internal metadata (`AtpMetadata`/`OsmMetadata`, via the same pipeline-level re-exports `atp`/`osm` already had). Nothing outside `pipeline` ever used it. Not really a top-level crate concern the way `geometry`/`matchers`/`tables` are — those get used across multiple otherwise-independent parts of the crate; `provenance` never did.

Pure move: `provenance.rs`'s own imports (`crate::pipeline::{self, AtpMetadata, OsmMetadata}`) needed no changes at all, since those are absolute paths unaffected by where the importing file itself lives. Only `lib.rs` (drop the top-level `mod`), `pipeline/mod.rs` (add the submodule), the one real call site in `pipeline::conflate::mod.rs`, and a handful of doc-comment cross-references elsewhere needed updating.

## Verification

`cargo build`/`test`/`clippy --all-targets`/`fmt --check` all clean, including the integration test, which exercises the real BOM-embedding path.


brawer (commented):
LGTM

PR #703, opened 2026-08-18
Move memstats/logging/s2_util into pipeline

brawer:
All three were only ever used from within `pipeline` — checked real call sites (`crate::memstats`/`crate::logging`/`crate::s2_util`), not just guessed:

- `memstats` (419 lines): only `pipeline/mod.rs`, `pipeline/osm/mod.rs`, `pipeline/conflate/mod.rs`.
- `logging` (237 lines): only `pipeline/mod.rs`.
- `s2_util`'s `MergedCellRanges`: only `pipeline/conflate/mod.rs`, and nowhere else in `pipeline` either — so it becomes `conflate`'s own private submodule (`pipeline::conflate::s2_util`), not a pipeline-wide one, matching how `conflate` already has its own private `writer` submodule.

Considered making them submodules of `utils` instead (that was the original question), but `utils.rs`'s own real content (`UtcTimestamp`, `to_hex`) is genuinely small and genuinely cross-cutting — used by `places::Place` *and* `pipeline::conflate::writer`, two otherwise-independent parts of the crate. `memstats`/`logging` are the opposite: substantial, self-contained, single-purpose subsystems (platform-specific cgroup introspection; a whole structured-logging setup) that just happen to only ever get called from `pipeline`. That's exactly the shape `provenance`/`geostats` had, so this follows that same move rather than merging into `utils` and diluting what "utils" means here. `utils.rs` itself doesn't need to become a directory for any of this.

## Verification

`cargo build`/`test`/`clippy --all-targets`/`fmt --check` all clean. Also ran the real pipeline against the test fixtures end to end: `pipeline.log` got created (`logging::init`), 16 memstats snapshots logged (`rss_bytes` present), and `conflated.parquet` was produced correctly (exercises `s2_util::MergedCellRanges` in the real OSM-candidate query path).


src/utils.rs
```diff
@@ -1,7 +1,12 @@
-//! Small, generic helpers shared by more than one place in the crate.
-//! A natural home for other such helpers as they accumulate -- e.g.
-//! `crate::memstats` is arguably one of these too, just not moved here
-//! (yet).
+//! Small, generic helpers shared by more than one otherwise-independent
+//! part of the crate -- e.g. `UtcTimestamp` below is used by both
+//! `places::Place::fetched` and `pipeline::conflate::writer`. A natural
+//! home for other such helpers as they accumulate, as long as they're
+//! genuinely small and genuinely cross-cutting: `pipeline::memstats`/
```
brawer: Stop after "cross-cutting.", no need to describe the history here.

brawer: Done in 5d1b443.

brawer (commented):
LGTM

Issue #704, opened 2026-08-18
Investigate unused modified() accessors: possible gap in pipeline memoization


brawer:
Several table types expose a `modified()` accessor (returning the mtime of
their backing file(s)), meant to support the pipeline's per-step
memoization: each stage compares its inputs' and its own output's
modification times to decide whether it needs to re-run. Right now, these
are never called anywhere in the codebase (only from their own unit
tests):

- `CoordTable::modified()` (`src/tables/coord_table.rs`)
- `GraphTable::modified()` (`src/tables/graph.rs`)
- `U64Set::modified()` (`src/tables/u64_set.rs`)
- `BlobTable::modified()` (`src/tables/blob_table.rs`)
- `GeometryTable::modified()` (`src/tables/geometry_table.rs`)

By contrast, `StringPool::modified()` and `OsmFeatureIndex::modified()`
*are* called for real, from `conflate()`'s staleness check
(`src/pipeline/conflate/mod.rs`) and `OsmFeatures::modified()`
(`src/tables/feature_index.rs`).

That five table types' `modified()` are dead suggests some pipeline
stage's cache-skip check isn't actually consulting one of its real
inputs' modification time -- i.e. a stage could see a newer input (e.g. a
changed `CoordTable`/`GraphTable`/`U64Set`/`BlobTable`/`GeometryTable` on
disk) and incorrectly decide its cached output is still fresh, silently
reusing stale data instead of recomputing.

This is **not a production concern**: production runs always start from
an empty working directory, so there's nothing to memoize against yet.
It's a correctness gap for local/developer workflows, where a workdir is
reused across runs specifically to skip already-done steps -- exactly the
scenario this memoization exists for.

**To investigate**: trace which pipeline stage(s) own each of these
tables as an *input* (not just as their own output) and confirm whether
that stage's staleness check (typically a `last_modified(&[...])?`
comparison, see `src/pipeline/mod.rs`) actually folds in that table's
`modified()`. Fix the check if it's missing, or -- if a table genuinely
never needs its own staleness tracked (e.g. it's always rebuilt alongside
something else whose mtime already covers it) -- document why and remove
the dead accessor instead.

Each of the five `modified()` methods currently carries `#[allow(unused)]`
with a comment pointing at this issue.

PR #705, opened 2026-08-18
Clean up #[allow(unused)]: drop stale ones, delete dead code, keep modified()

brawer:
Audited all 30 `#[allow(unused)]` in the codebase. Manual grepping for call sites turned out unreliable on its own (missed that a few module-level `#[allow(unused)]` on `mod` declarations were blanket-suppressing warnings for everything inside those files, and separately conflated a `strings.len()` call on `StringCounts` with one on `StringPool`) -- so this is based on temporarily stripping every attribute and reading `cargo check`'s actual dead-code output, not just grep.

**Stale attributes removed** -- item is genuinely used elsewhere, just deleting the now-pointless annotation: `ConflatedFeature`, `Assembly`, `Prunings`, `leaf_relations_geometry`, `StringPool::get`/`lookup`, and the module-level allows on `BlobTable`/`GeometryStore`/`RecordReader`.

**Dead code deleted** (and its test coverage, where a test existed solely to exercise it) -- no real caller anywhere, including tests:
- `Prunings::coord`/`keep_way`/`keep_relation` (callers use the fields directly instead)
- `BlobTable::iter`/`GeometryTable::iter`
- `GeometryStore::contains_key`/`len`/`is_empty`
- `RecordReader::is_empty`
- `StringPool::len` (and its now-unread backing field) -- notably, every *sibling* table's `len()` is genuinely used in production; this one only had test coverage.

**Kept, not deleted**: `modified()` on `CoordTable`/`BlobTable`/`GraphTable`/`U64Set`/`GeometryTable`. These back the pipeline's per-step memoization (staleness checks comparing input/output mtimes), and them being unused looks like a real gap in that memoization rather than dead code -- filed #704 to investigate, linked from each `#[allow(unused)]`.

**Left untouched**: `OsmFeatureIndex`'s `entries_count`/`id`/`len`/`is_empty`/`feature_id` and the `LocalFeatureRef` re-export -- explicitly documented in-code as scaffolding for `suggest_edits`' future move onto this index.

Verified: `cargo build`/`test`/`clippy`/`fmt` clean (245 lib + 2 bin + 4 integration tests).

### DIFF
--- a/src/pipeline/conflate/mod.rs
+++ b/src/pipeline/conflate/mod.rs
@@ -77,7 +77,6 @@ pub fn conflate(
     Ok(out_path)
 }
 
-#[allow(unused)]
 struct ConflatedFeature {
     atp: Option<Place>,
     osm: Option<Feature>,
--- a/src/pipeline/osm/assemble.rs
+++ b/src/pipeline/osm/assemble.rs
@@ -29,7 +29,6 @@ use std::{
 };
 use wkb::writer::{write_geometry, write_point};
 
-#[allow(unused)]
 pub struct Assembly<'a> {
     pub strings: StringPool<'a>,
     pub nodes: RecordReader,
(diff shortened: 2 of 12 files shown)


brawer (commented):
LGTM

PR #710, opened 2026-08-18
edits.rs: fix dangling reference to a since-removed Place::to_geojson

brawer:
Found while auditing for stale/undocumented loose ends before the v0.7.0 release.

`SuggestedEdit`'s doc comment pointed at "the TODO on `crate::places::Place::to_geojson`" — that method doesn't exist anymore (`to_geojson` lives on `SuggestedEdit` itself now, per #655's pipeline rewrite). Just states the constraint directly instead of pointing at a comparison that no longer has a target.


brawer (commented):
LGTM

PR #717, opened 2026-08-19
Bump s2 to 0.2.0, fixing RUSTSEC-2026-0196/0197 (cgmath)

brawer:
s2 0.2.0 drops its dependency on cgmath entirely — the small Matrix3-based
"frame" abstraction (used for orthonormal-basis conversions) was replaced
with dot products against s2's own `Vector` type. See
[yjh0502/rust-s2#115](https://github.com/yjh0502/rust-s2/issues/115).

This removes `cgmath`, `approx`, `float_extras`, and `lazy_static` from the
dependency tree and clears both cgmath RUSTSEC advisories for good, rather
than just leaving them unreachable (`swap_columns` — the buggy API — was
never actually called by s2 or by us).

Pure dependency bump, no source changes needed on our side (`cargo update -p s2`
+ Cargo.toml version bump). Full test suite (245 unit + 4 integration tests)
passes, including S2-geometry-heavy pipeline code — a good real-world check
that the rewrite in s2 0.2.0 works correctly, not just that it compiles.

https://osv.dev/RUSTSEC-2026-0196
https://osv.dev/RUSTSEC-2026-0197

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -75,15 +75,6 @@ checksum = "330a5ed07fa54e4702c9d6c4174f74427fc0ef6e214bbd677ae50a5099946470"
 
 [[package]]
 name = "approx"
-version = "0.4.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


brawer (commented):
LGTM

Issue #730, opened 2026-08-20
Remove changeset from our pipeline


brawer:
## How this came up

While working on a script for smoke-testing released containers on real cloud hardware, we ran the `v0.7.1` pipeline against a regional excerpt for Switzerland, downloaded from Geofabrik, instead of the full OSM planet — to make iteration cheap. The pipeline panicked partway through `conflate`. The same pipeline had processed the full OSM planet via BitTorrent just fine, on the same version, shortly before. See #729 for the technical debugging trail; this issue is about what we found while chasing down *why* the two runs differed, not the crash itself.

## What we found: Geofabrik removes changeset (and user/uid) on purpose

The Geofabrik excerpt lacked `changeset`, `user`, and `uid` for every object — not a one-off gap. [Geofabrik's own download page](https://www.geofabrik.de/data/download.html) explains why, in the "Rohdaten mit personenbezogenen Metadaten und voller Versionsgeschichte" ("Raw data with personally-identifiable metadata and full version history") section:

> Zu den OpenStreetMap-Rohdaten gehören in der Regel auch Metadaten. Einige der Metadatenfelder haben offensichtlichen Personenbezug (Benutzername, Benutzernummer, Änderungssatznummer). Die Rohdatendateien (.osm.pbf, .osc.gz, .osm.bz2) auf dem öffentlichen Downloadserver (download.geofabrik.de) enthalten diese Felder nicht. Unsere kostenlosen Shapefiles enthalten keinerlei Metadaten.
>
> Da die Metadatenfelder nur von den OpenStreetMap-Mitwirkenden selbst zu Analysezwecken (z.B. Vandalismusbekämpfung) benötigt werden, bieten wir auf einem separaten Downloadserver unter osm-internal.download.geofabrik.de passwortgeschützt Rohdatensätze mit allen Metadatenfeldern sowie regionale Auszüge mit voller Versionsgeschichte an. Dieses kostenlose Angebot kann nur von OpenStreetMap-Mitwirkenden genutzt werden.

("OpenStreetMap raw data usually includes metadata too. Some of the metadata fields have an obvious connection to a person (username, user number, changeset number). The raw data files ... on the public download server (download.geofabrik.de) do not contain these fields. ... we offer raw datasets with all metadata fields ... password-protected on a separate download server ... only for OpenStreetMap contributors.")

`changeset` — `Änderungssatznummer` — is named explicitly, alongside username and user ID, as one of the fields "with an obvious connection to a person."

[...]

PR #731, opened 2026-08-20
fix!: remove OSM changeset from pipeline output

brawer:
## What

Removes `changeset` from `conflated.parquet`'s output schema (`osm.modified.changeset`) and from the `@osm_changeset` GeoJSON property in suggested-edits output, along with all the internal plumbing that carries it (`tables::features::Feature`, `conflate::writer`'s Arrow builders/schema, `edits::SuggestedEdit`).

## Why

Full reasoning in #730. Short version: a changeset ID resolves, via OSM's own public API, to the account that made the edit — unlike a node/way/relation ID, which only identifies a place. Geofabrik excludes exactly this field from their public downloads for this reason, citing EU data protection law. Our one concrete downstream consumer, MapRoulette, doesn't need it either (task identity is by OSM type+id, not changeset).

`version`/`timestamp` are unaffected — neither is personal data, and `version` is what OSM's own edit API actually uses for conflict detection, which is what `@osm_changeset` was there for in the first place.

## Is this really `!`?

Yes, by this repo's own rule (`docs/CONTRIBUTING.md`): any output-schema shape change is `!`, regardless of whether it's additive/removal/rename, and regardless of whether anything currently depends on it. That said, worth having on record per #730: this pipeline isn't in production yet, so there's no real downstream consumer to actually break today.

## Testing

- `cargo fmt` / `cargo clippy --all-targets -- -D warnings` / `cargo test` all pass — 246 unit tests + 4 integration tests, including `test_pipeline`'s real fixture-based run against `tests/test_data/zugerland.osm.pbf`.
- Updated `tests/integration_test.rs`'s assertions accordingly, including turning the old `@osm_changeset` presence check into a regression check that it's now *absent* (`feature["properties"].get("@osm_changeset").is_none()`), so this can't silently regress later.
- Updated `docs/outputs/CONFLATED_PARQUET.md`'s schema table and added a note on the intentional omission, linking back to #730.

## Test data files: no real changeset/PII either

Worth having on record, since this repo's test fixtures also carry OSM metadata: neither checked-in fixture contains real changeset (or username/uid) data.

- **`tests/test_data/zugerland.osm.pbf`** (the fixture `test_pipeline` and the Rust unit tests actually exercise): `osmium fileinfo -e` reports `Number of changesets: 0` and `Smallest/Largest changeset ID: 0` — every element's changeset field is unset in this extract.
- **`tests/test_data/osm-testdata-grid/`** (used only by `src/geometry/grid_tests.rs`, unrelated to the conflation pipeline this PR touches): its `.osm` XML files *do* carry `changeset`/`uid`/`user` attributes, but they're synthetic placeholders from the OSM community's own upstream [`osm-testdata`](https://github.com/osmcode/osm-testdata) conformance-test project (vendored via `scripts/vendor-osm-testdata-grid.sh`) — e.g. `changeset="1" uid="1" user="test" generator="testdata" upload="false"`. One fixture even carries a deliberately 250+ character fake username, clearly a length-edge-case stress value, not a real person.

### DIFF
--- a/docs/outputs/CONFLATED_PARQUET.md
+++ b/docs/outputs/CONFLATED_PARQUET.md
@@ -30,7 +30,6 @@ osm                                          STRUCT, nullable
   tags        MAP<UTF8, UTF8>                non-null map, non-null keys and values
   modified                                   STRUCT, non-null
     timestamp TIMESTAMP(ms, UTC)             non-null
-    changeset UINT64                         non-null
     version   UINT32                         non-null
   way_members LIST<UINT64>                   nullable — present only when type = "way"
   relation_members LIST<STRUCT>              nullable — present only when type = "relation"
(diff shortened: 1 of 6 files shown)


docs/outputs/CONFLATED_PARQUET.md
```diff
@@ -57,10 +56,17 @@ A few things worth calling out that aren’t obvious from the above:
...
-  preserve OpenStreetMap’s full edit history here, only its current
-  state as of the database snapshot this file was built from.
+  OpenStreetMap’s own edit metadata (the timestamp and version number
+  of that feature’s most recent edit). We don’t preserve OpenStreetMap’s
+  full edit history here, only its current state as of the database
+  snapshot this file was built from.
+- **We deliberately don’t carry OpenStreetMap’s changeset ID.** A
+  changeset number resolves, via OSM’s own public API, to the account
+  that made the edit — unlike a node/way/relation ID, which only
+  identifies a place. Our downstream use case ([MapRoulette](https://maproulette.org/))
```
brawer: Remove "Our downstream use case ([MapRoulette](https://maproulette.org/)) doesn’t need it either." - While correct, I think this is too much detail for this schema description. If people want to know more, they can click on the link to #730 for the full analysis and trade-offs, which already states (in more detail) what MapRoulette needs.

